PR #80, opened 2018-04-25
[79] Set deployment target to 10.0 and update project settings

### LINKED ISSUE #79: Lower deployment target to 10.0
I think a policy of always supporting the current major iOS version - 1 would be nice to support  "legacy" projects. 😉 

The current deployment target is 10.3 so I suggest lowering it to 10.0.

This shouldn't cause any problems.

daniel-mohemian:
Resolves #79.

- Lower the deployment target from 10.3 to 10.0
- Update to recommended project settings
- Resolve an _overlapping access_ warning


JOSESwift/Sources/CryptoImplementation/SecureRandom.swift
```diff
@@ -38,7 +38,7 @@ internal struct SecureRandom {
         var generatedRandom = Data(count: count)
 
         let randomGenerationStatus = generatedRandom.withUnsafeMutableBytes { mutableRandomBytes in
-            SecRandomCopyBytes(kSecRandomDefault, generatedRandom.count, mutableRandomBytes)
+            SecRandomCopyBytes(kSecRandomDefault, count, mutableRandomBytes)
```
gigi-mohemian: 👍

PR #83, opened 2018-05-21
[83] Add support for RS256

turbulem:
Hi, I've added support for JWS RS256 algorithm. It's trivial and is needed for my project.
Let me know if you have any questions.

Changes:
- Support for encoding/decoding JWS RS256
- Tests
- Updated Readme.md

PR #84, opened 2018-05-22
Prepare 1.1.0

daniel-mohemian:
We should really automate this (#85). 🚀 

**Changes:**

- Add RS256 algorithm for JWS (#83)


JOSESwift/Sources/Signer.swift
```diff
@@ -45,7 +45,7 @@ public struct Signer<KeyType> {
     /// - Returns: A fully initialized `Signer` or `nil` if provided key is of the wrong type.
     public init?(signingAlgorithm: SignatureAlgorithm, privateKey: KeyType) {
         switch signingAlgorithm {
-        case .RS256,.RS512:
+        case .RS256, .RS512:
```
gigi-mohemian: 😄👍

PR #94, opened 2018-07-18
Update plain text length check for RSA1_5

garrefa:
https://tools.ietf.org/html/rfc3447#section-7.2.1

M message to be encrypted, an octet string of length mLen, **where mLen <= k - 11**

It was probably a mistake caused by the step 1 description:

   1. Length checking: If mLen > k - 11, output "message too long" and
      stop.


gigi-mohemian:
> In your extended test suite, it would be awesome if you could include some edge case testing for such things as maximum plaintext length etc. 

I will, I promise! 😅

garrefa:
@gigi-mohemian great man, I saw there was no tests for it and was going to add in a next MR. Im trying to add support for RSAES-OAEP and will add tests for sure.

Issue #93, opened 2018-07-17
Update documentation


daniel-mohemian:
Would be cool if we could update the wiki for

- [ ] Direct encryption
- [ ] Symmetric JWKs
- [ ] JWK Sets

Soon also for:

- [ ] EC JWKs
- [ ] EC JWS

PR #97, opened 2018-07-30
[67] Add option to decrypt and verify with explicit decrypters and verifiers

### LINKED ISSUE #67: Let the user provide an explicit verification algorithm instead of inferring it from the header
See [this article](https://auth0.com/blog/critical-vulnerabilities-in-json-web-token-libraries/) by Auth0. Possibly also applies to JWE decryption.

daniel-mohemian:
Resolves #67. This should help protect against [this attack](https://auth0.com/blog/critical-vulnerabilities-in-json-web-token-libraries/) where an attacker changes the algorithm in the header of a JWS, forcing libraries to validate the JWS signature using a wrong algorithm.

Now the user can explicitly state which algorithm is used to verify the signature by providing a `Verifier` not just a key.

For consistency, I added the same for JWE decryption.

### JWS

The following validation functions got deprecated

``` swift
func isValid<KeyType>(for publicKey: KeyType) -> Bool
func validate<KeyType>(with publicKey: KeyType) throws -> JWS
```

in favor of

``` swift
func isValid(for verifier: Verifier) -> Bool
func validate(using verifier: Verifier) throws -> JWS
```

### JWE

The following decryption function got deprecated

``` swift
func decrypt<KeyType>(with key: KeyType) throws -> Payload
```

in favor of

``` swift
func decrypt(using decrypter: Decrypter) throws -> Payload
```

I updated the documentation in the readme accordingly. The wiki needs to be updated. See #93.


JOSESwift/Sources/JWS.swift
```diff
@@ -152,6 +154,44 @@ public struct JWS {
...
         return self
     }
+
+    /// Checks whether the JWS's signature is valid using a given verifier.
+    ///
+    /// - Parameter verifier: The verifier containing the public key whose corresponding private key signed the JWS.
+    /// - Returns: The JWS on which this function was called if the signature is valid.
+    /// - Throws: A `JOSESwiftError` if the signature is invalid or if errors occured during signature validation.
+    public func validate(using verifier: Verifier) throws -> JWS {
+        guard verifier.verifier.algorithm == header.algorithm else {
```
carol-mohemian: This looks really strange imo. Could we maybe rename the "verifierProtocol" to something else? Can't come up with a better name though.

Tests/JWEDirectEncryptionTests.swift
```diff
@@ -43,7 +43,8 @@ class JWEDirectEncryptionTests: CryptoTestCase {
         121,  25,   4,  85, 220, 144, 215, 110,
         130,  17,  68, 228, 129, 138,   7, 130
     ])
-    
+
+    @available(*, deprecated)
```
carol-mohemian: Hmm, what's the plan for all those deprecated tests? Do you write new ones?

daniel-mohemian: No, they use a deprecated method (the decrypt method; which should be fine for the test since we're not attacking ourselves 😉). If I don't mark them deprecated we get a warning for each decrypt call. If we want to get rid of the annotations, we just have to use the other decrypt method (the one which takes a verifier).

I can update that now if you want?

carol-mohemian: That's what I meant. But it's fine for me.

JOSESwift/Sources/JWS.swift
daniel-mohemian: I currently can only think of `Verifier` and `Validator` but that will be strange too, having two different names (although I do like validator more than verifier somehow 😅). We could rename the parameter in the validate method. But to what? 😅

Tests/JWERSATests.swift
```diff
@@ -81,4 +87,85 @@ class JWETests: CryptoTestCase {
...
+            contentDecryptionAlgorithm: .A256CBCHS512
+            )!
+
+        XCTAssertThrowsError(try jwe.decrypt(using: decrypter), "decrypting with wrong header alg") { error in
+            XCTAssertEqual(error as! JOSESwiftError, JOSESwiftError.decryptingFailed(description: "JWE header algorithms do not match encrypter algorithms."))
+        }
+    }
+
+    func testDecryptWithExplicitDecrypterWrongEncInHeader() {
+        // Replaces alg "A256CBC-HS512" with alg "A128GCM" in header
```
gigi-mohemian: I think this should be `enc` instead of `alg` both times, right?

Tests/JWERSATests.swift
```diff
@@ -81,4 +87,85 @@ class JWETests: CryptoTestCase {
...
+            keyDecryptionAlgorithm: .RSA1_5,
+            decryptionKey: privateKeyAlice2048!,
+            contentDecryptionAlgorithm: .A256CBCHS512
+            )!
+
+        XCTAssertEqual(try! jwe.decrypt(using: decrypter).data(), plaintext)
+    }
+
+    func testDecryptWithExplicitDecrypterWrongAlgInHeader() {
+        // Replaces alg "RSA1_5" with alg "RSA-OAEP" in header
```
gigi-mohemian: 👍

Tests/JWERSATests.swift
```diff
@@ -81,4 +87,85 @@ class JWETests: CryptoTestCase {
...
+
+        let jwe = try! JWE(compactSerialization: malformedSerialization)
+
+        let decrypter = Decrypter(
+            keyDecryptionAlgorithm: .RSA1_5,
+            decryptionKey: privateKeyAlice2048!,
+            contentDecryptionAlgorithm: .A256CBCHS512
+            )!
+
+        XCTAssertThrowsError(try jwe.decrypt(using: decrypter), "decrypting with wrong header alg") { error in
```
gigi-mohemian: `with wrong header alg` -> `with wrong encryption algorithm in header`

daniel-mohemian:
Thanks for reviewing @gigi-mohemian. I addressed your feedback.

Issue #98, opened 2018-08-14
Fix swiftlint errors and warnings


daniel-mohemian:
So we can setup required linitng with danger and for release automation. Make sure to update the call in the Fastfile to `swiftlint(ignore_exit_status: false)`.

PR #100, opened 2018-08-22
Add basic release automation

daniel-mohemian:
A basic way to automate releases. Let's see how it holds up and improve it on the go I'd say.

## Release Process

- *`prepare`*: Runs bumps version, formats code, runs tests, and creates release preparation branch
- Pull request is merged
- *`release`*: Tags the release and publishes the pod

Let me know what you think. ⚡️


fastlane/Fastfile
```diff
@@ -0,0 +1,121 @@
...
+
+desc "Run the tests"
+lane :test do
+  scan(scheme: "JOSESwift")
+end
+
+desc "Lint Swift files"
+lane :lint do
+  # Todo: Don't ignore exit status (#98)
+  swiftlint(ignore_exit_status: true)
```
gigi-mohemian: Are we ok with ignoring the `swiftlint` exit status for now?

fastlane/Fastfile
```diff
@@ -0,0 +1,121 @@
...
+      new_version_xcode = increment_version_number
+      new_version_podspec = version_bump_podspec(path: "JOSESwift.podspec")
+    else
+      new_version_xcode = increment_version_number(version_number: version)
+      new_version_podspec = version_bump_podspec(path: "JOSESwift.podspec", version_number: version)
+    end
+  end
+
+  if new_version_xcode == new_version_podspec
+    UI.success "Did bump version number to #{new_version_xcode}"
```
gigi-mohemian: Maybe the wording could be improved here. `Did bump` -> `Bumped`

fastlane/Fastfile
```diff
@@ -0,0 +1,121 @@
...
+  # Ensure we prepare from current master
+  ensure_git_branch(branch: "master")
+  ensure_git_status_clean
+  sh("git", "pull")
+  ensure_git_status_clean
+
+  # Bump version
+  case bump_type
+  when "major", "minor", "patch"
+    version = bump(type: bump_type)
```
gigi-mohemian: You could do `version = bump(options)` and skip the whole `case` statement.

fastlane/Fastfile
```diff
@@ -0,0 +1,121 @@
...
+  end
+
+  new_version_xcode
+end
+
+desc "Prepares a release by bumping version, formatting code, running tests, setting up a release preparation branch"
+lane :prepare do |options|
+  bump_type = options[:type]
+  version = ""
+  branch_name = ""
```
gigi-mohemian: No need to initialize these three variables.

fastlane/Fastfile
```diff
@@ -0,0 +1,121 @@
...
+
+  # Ensure we prepare from current master
+  ensure_git_branch(branch: "master")
+  ensure_git_status_clean
+  sh("git", "pull")
```
gigi-mohemian: Is doing a `git pull` here necessary? If yes, why?

fastlane/Fastfile
daniel-mohemian: We have to, for now. Otherwise, we can't get past that step in the automation because we get a whole bunch of warnings errors currently. We have to get rid of them first and make a decision which ones to ignore. I setup #98 for that.

fastlane/Fastfile
daniel-mohemian: Not needed necessarily since the release branch itself will go into master again. 👍

daniel-mohemian:
Thanks @gigi-mohemian I addressed your feedback.

fastlane/Fastfile
```diff
@@ -0,0 +1,114 @@
...
+  end
+
+  new_version_xcode
+end
+
+desc "Prepares a release by bumping version, formatting code, running tests, setting up a release preparation branch"
+lane :prepare do |options|
+  bump_type = options[:type]
+  version
+  branch_name
```
gigi-mohemian: I think you can remove all three variables above completely now 👍

daniel-mohemian: Right! Feels like JS now. 😅

PR #110, opened 2018-10-02
Add setters for additional JOSEHeader parameters

### LINKED ISSUE #109: Initialize `JWEHeader` with additional parameters
At the moment the only way to initialize a `JWEHeader` is with the following syntax, like it is described in the [documentation](https://github.com/airsidemobile/JOSESwift/wiki/jwe#header-1): 

```swift 
 let header = JWEHeader(algorithm: .RSA1_5, encryptionAlgorithm: .A256CBCHS512)
```

It's not quite clear to me why all the parameters (see [here](https://github.com/airsidemobile/JOSESwift/blob/master/JOSESwift/Sources/JWEHeader.swift#L94)) are accessible to read but not to write. The public initializer described above is the only way to initialize the JWE header, so all the parameters have to be always `nil`, right? 

I would be glad, if you could give me a hint, if it's possible to declare an additional parameter like `kid` and how to do it.

fbernutz:
Fixes #109 

## Description 

Add an additional argument for the `kid` parameter (defaults to nil) when initializing `JWEHeader`. So it can be called with a `kid` parameter or easily without. So this is not a breaking change.

``` swift
// With kid
JWEHeader(algorithm: .RSA1_5, encryptionAlgorithm: .A256CBCHS512, keyIdentifier: "kid")

// Without kid
JWEHeader(algorithm: .RSA1_5, encryptionAlgorithm: .A256CBCHS512)
```

The additional parameter gets set to the internal `parameter` dictionary, so when reading out the value of `kid` the value gets returned. 

``` swift
let header = JWEHeader(algorithm: .RSA1_5, encryptionAlgorithm: .A256CBCHS512, keyIdentifier: "kid")
print(header.kid) // "kid"
```

## Tests

I added a test for this.

## Update Documentation

I am not quite sure how the contribution-flow should look like for editing the Wiki as I can't put the changes of the Wiki here into this PR. So, I updated the README and added the `keyIdentifier` parameter when initializing the `JWEHeader`. But, I think, an extra information in the wiki would be nice. I opened a new issue #111 with updated content for the wiki, you can then use the content from there to update the wiki.

### DIFF
--- a/JOSESwift/Sources/JWEHeader.swift
+++ b/JOSESwift/Sources/JWEHeader.swift
@@ -54,7 +54,8 @@ public struct JWEHeader: JOSEHeader {
         self.parameters = parameters
     }
 
-    /// Initializes a `JWEHeader` with the specified algorithm and signing algorithm.
+    /// Initializes a `JWEHeader` with the specified algorithm, signing algorithm
+    /// and optionally a key identifier.
     public init(algorithm: AsymmetricKeyAlgorithm, encryptionAlgorithm: SymmetricKeyAlgorithm, keyIdentifier: String? = nil) {
         var parameters = [
             "alg": algorithm.rawValue,
--- a/JOSESwift/Sources/JWEHeader.swift
+++ b/JOSESwift/Sources/JWEHeader.swift
@@ -102,10 +102,13 @@ extension JWEHeader: CommonHeaderParameterSpace {
     /// one of which corresponds to the key used to encrypt the JWE.
     public var jku: URL? {
         set {
-            parameters["jku"] = newValue
+            parameters["jku"] = newValue?.absoluteString
         }
         get {
-            return parameters["jku"] as? URL
+            guard let parameter = parameters["jku"] as? String else {
+                return nil
+            }
+            return URL(string: parameter)
         }
     }
 
@@ -133,10 +136,13 @@ extension JWEHeader: CommonHeaderParameterSpace {
     /// or certificate chain corresponding to the key used to encrypt the JWE.
     public var x5u: URL? {
         set {
-            parameters["x5u"] = newValue
+            parameters["x5u"] = newValue?.absoluteString
         }
         get {
-            return parameters["x5u"] as? URL
+            guard let parameter = parameters["x5u"] as? String else {
+                return nil
+            }
+            return URL(string: parameter)
         }
     }
 
--- a/JOSESwift/Sources/JWEHeader.swift
+++ b/JOSESwift/Sources/JWEHeader.swift
@@ -54,17 +54,12 @@ public struct JWEHeader: JOSEHeader {
         self.parameters = parameters
     }
 
-    /// Initializes a `JWEHeader` with the specified algorithm, signing algorithm
-    /// and optionally a key identifier.
-    public init(algorithm: AsymmetricKeyAlgorithm, encryptionAlgorithm: SymmetricKeyAlgorithm, keyIdentifier: String? = nil) {
-        var parameters = [
+    /// Initializes a `JWEHeader` with the specified algorithm and signing algorithm.
+    public init(algorithm: AsymmetricKeyAlgorithm, encryptionAlgorithm: SymmetricKeyAlgorithm) {
+        let parameters = [
             "alg": algorithm.rawValue,
             "enc": encryptionAlgorithm.rawValue
         ]
-
-        if let keyIdentifier = keyIdentifier {
-            parameters["kid"] = keyIdentifier
-        }
 
         // Forcing the try is ok here, since [String: String] can be converted to JSON.
         // swiftlint:disable:next force_try
--- a/JOSESwift/Sources/JWEHeader.swift
+++ b/JOSESwift/Sources/JWEHeader.swift
@@ -28,7 +28,10 @@ public struct JWEHeader: JOSEHeader {
     var headerData: Data
     var parameters: [String: Any] {
         didSet {
-            // Forcing the try is ok here, since [String: String] can be converted to JSON.
+            guard JSONSerialization.isValidJSONObject(parameters) else {
+                return
+            }
+            // Forcing the try is ok here, because it is valid JSON.
             // swiftlint:disable:next force_try
             headerData = try! JSONSerialization.data(withJSONObject: parameters, options: [])
         }
--- a/JOSESwift/Sources/JWEHeader.swift
+++ b/JOSESwift/Sources/JWEHeader.swift
@@ -142,12 +142,12 @@ extension JWEHeader: CommonHeaderParameterSpace {
 
     /// The X.509 certificate chain contains the X.509 public key certificate or
     /// certificate chain corresponding to the key used to encrypt the JWE.
-    public var x5c: [String: Any]? {
+    public var x5c: [String]? {
         set {
             parameters["x5c"] = newValue
         }
         get {
-            return parameters["x5c"] as? [String: Any]
+            return parameters["x5c"] as? [String]
         }
     }
 
--- a/JOSESwift/Sources/JWEHeader.swift
+++ b/JOSESwift/Sources/JWEHeader.swift
@@ -25,8 +25,14 @@ import Foundation
 
 /// The header of a `JWE` object.
 public struct JWEHeader: JOSEHeader {
-    let headerData: Data
-    let parameters: [String: Any]
+    var headerData: Data
+    var parameters: [String: Any] {
+        didSet {
+            // Forcing the try is ok here, since [String: String] can be converted to JSON.
+            // swiftlint:disable:next force_try
+            headerData = try! JSONSerialization.data(withJSONObject: parameters, options: [])
+        }
+    }
 
     /// Initializes a JWE header with given parameters and their original `Data` representation.
     /// Note that this (base64-url decoded) `Data` representation has to be exactly as it was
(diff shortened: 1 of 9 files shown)


daniel-mohemian:
Thanks so much, @fbernutz for the contribution! 🎉 

We will have a look at it this week, latest Thursday! Hope that's fine with you.

fbernutz:
I'm happy that this project is open source and that I can contribute 😄 Thanks for your answer, I'm fine with a review the next days!

carol-mohemian:
Hey @fbernutz 👋 and thanks for your contribution also from my side! 😍 

@daniel-mohemian and I discussed your approach and we came to the conclusion that a "more generic" approach would be nice, mainly to avoid having an initializer with all the (optional) "additional parameters". 

To explain our approach I want to give you some background on how and why we implemented the parameter handling as it is at the moment. If you look at the [`init(algorithm: AsymmetricKeyAlgorithm, encryptionAlgorithm: SymmetricKeyAlgorithm)` in JWEHeader.swift](https://github.com/airsidemobile/JOSESwift/blob/eb339b25ad545da85b1896960afbfa2d7b5d840e/JOSESwift/Sources/JWEHeader.swift#L58) you can see that the `headerData` is serialized from the `parameters` dictionary. This is because every time the dictionary is passed to another method the order of the dictionary can change. For the decrypting/verifying part it's crucial that this data doesn't change at all. 

Therefore, we would suggest that you change your implementation a little:
If you look at the [`CommonHeaderSpace` protocol](https://github.com/airsidemobile/JOSESwift/blob/eb339b25ad545da85b1896960afbfa2d7b5d840e/JOSESwift/Sources/JOSEHeader.swift#L64) you can see that the "additional parameters" are get-only properties at the moment. You could add a `set` there as well. 

Coming to the tricky part: in the protocol implementation set the value of the parameter in the parameter dictionary to the actual `newValue` of the property. The parameter property of the `JWEHeader` needs to be mutable, of course. To update the `headerData` (also mutable) you need to "recompute" the data from the new parameters dictionary by serializing it again. 
If all this is done you can set your desired parameter as shown below:

```
var header = JWEHeader(algorithm: .RSA1_5, encryptionAlgorithm: .A256CBCHS512)
header.kid = "bla"
let payload = Payload(message.data(using: .utf8)!)
let encrypter = Encrypter(keyEncryptionAlgorithm: .RSA1_5, encryptionKey: publicKeyAlice2048!, contentEncyptionAlgorithm: .A256CBCHS512)!
```

Hopefully this is not too much of a bummer for you since you did already a great job implementing the initializer with the optional parameter. 
If you have any question or if you are not convinced of our approach don't hesitate to ask question or suggesting another approach.

fbernutz:
Thank you so much for your feedback! 🙂I think I understand what you explained and I already had a look at it. 
Just one question, I am not sure when to exactly update the `headerData`. I currently use a `didSet` on the `parameter` property, so the `headerData` gets updated every time a parameter is set. 
Is this a solution which works for you? You can have a look at my latest commits.

JOSESwift/Sources/JWEHeader.swift
```diff
@@ -25,8 +25,14 @@ import Foundation
...
 /// The header of a `JWE` object.
 public struct JWEHeader: JOSEHeader {
-    let headerData: Data
-    let parameters: [String: Any]
+    var headerData: Data
+    var parameters: [String: Any] {
+        didSet {
+            // Forcing the try is ok here, since [String: String] can be converted to JSON.
+            // swiftlint:disable:next force_try
+            headerData = try! JSONSerialization.data(withJSONObject: parameters, options: [])
```
fbernutz: I am not sure when to exactly update the `headerData`. I currently use a `didSet` on the `parameter` property, so the `headerData` gets updated every time a parameter is set. Is this a solution which works for you?

Tests/JWEHeaderTests.swift
```diff
@@ -125,4 +125,61 @@ class JWEHeaderTests: XCTestCase {
...
+        XCTAssertEqual(header.parameters["jwk"] as? String, jwk)
+        XCTAssertEqual(header.jwk, jwk)
+
+        XCTAssertEqual(header.parameters["kid"] as? String, kid)
+        XCTAssertEqual(header.kid, kid)
+
+        XCTAssertEqual(header.parameters["x5u"] as? URL, x5u)
+        XCTAssertEqual(header.x5u, x5u)
+
+//        XCTAssertEqual(header.parameters["x5c"] as? [String: Any], x5c)
```
fbernutz: I don't get these two commented lines to work, maybe you can have a look at them?

JOSESwift/Sources/JWEHeader.swift
carol-mohemian: I think this is perfectly fine 👌 😍

Tests/JWEHeaderTests.swift
carol-mohemian: This is probably because Dictionaries are not conforming to the `Equatable` protocol.

carol-mohemian:
This looks really awesome! Good job 🎉 
Regarding the CI: Builds on my machine as well without any problems. Can you try setting the Xcode version to 10 in the `.travis.yml` please? Thanks in advance 🙏

Regarding the "non-equatable" dicts you could probably add a custom `==` operator like shown in [this SO answer](https://stackoverflow.com/a/32366918/6825375). 

One more thing: Maybe add 1 more tests to the JWE and JWS tests where you do a roundtrip (encrypting/decrypting, signing/verifying) with a header where a "not required" parameters was set after the header was initialized, and see if everything works out.

JOSESwift/Sources/JWSHeader.swift
```diff
@@ -79,56 +85,105 @@ extension JWSHeader {
 extension JWSHeader: CommonHeaderParameterSpace {
     /// The JWK Set URL which refers to a resource for a set of JSON-encoded public keys,
     /// one of which corresponds to the key used to sign the JWS.
-    public var jku: URL? {
-        return parameters["jku"] as? URL
+    public var jku: URL? {set {
```
carol-mohemian: Formating

fbernutz: Ups, thanks for the hint ;)

JOSESwift/Sources/JWEHeader.swift
daniel-mohemian: Hi @fbernutz, 

that's a perfect place to do the data computation.

Just one thing: a `[String: Any]` dictionary that contains an URL won't be JSON encodable. The `try!` will fail here if the `parameters` contain a URL. What would you think about converting the URLs to `String` before setting them in the parameters dictionary and converting the `String` back to a `URL` when getting the parameter. E.g. like this:

``` swift
public var jku: URL? {
  set {
    parameters["jku"] = newValue.absoluteString
  }
  get {
    return URL(string: parameters["jku"])
   }
}	    
```

JOSESwift/Sources/JWEHeader.swift
```diff
@@ -95,55 +101,105 @@ extension JWEHeader: CommonHeaderParameterSpace {
...
+            parameters["x5u"] = newValue
+        }
+        get {
+            return parameters["x5u"] as? URL
+        }
     }
 
     /// The X.509 certificate chain contains the X.509 public key certificate or
     /// certificate chain corresponding to the key used to encrypt the JWE.
     public var x5c: [String: Any]? {
```
daniel-mohemian: Could you please change this parameter to `[String]` as per  [RFC-7515](https://tools.ietf.org/html/rfc7515#section-4.1.6)? Then we'll only have `[String]` and `String` parameters and the encoding to JSON data will work just fine. 😄

Tests/JWEHeaderTests.swift
daniel-mohemian: As noted above, you can change the type of `x5c` to `[String]`. This should solve this problem.

daniel-mohemian:
Very cool @fbernutz!

The Xcode10 change seems to have done the trick for the CI tests. 👍 

I added two comments regarding the types of the parameters. Please have a look at them and tell me what you think. Afterwards, we're ready to merge I'd say. 🎉

Tests/JWEHeaderTests.swift
fbernutz: @daniel-mohemian Perfect, with `[String]` this is no problem anymore ;)

JOSESwift/Sources/JWEHeader.swift
fbernutz: Hi @daniel-mohemian 🙂 
What should happen if it is somehow not a valid JSON? I thought about throwing an error like this: `HeaderParsingError.headerIsNotValidJSONObject`. But who could catch the thrown error? I am not quite sure from where the `headerData` is read. I could just return and don't update the headerData. What do you think about it?

fbernutz:
Hi @daniel-mohemian and @carol-mohemian,
I am almost finished with working on the feedback you gave me - thanks for that again! 😉
And a yay for the now again succeeding tests on the CI 🎉 

I just wanted to give you a little summary of what is still to do. 
Two issues still needs to be done: 
- As pointed out from Carol, I still want to write a test for JWE and JWS with the whole roundtrip of encrypting/decrypting, signing/verifying and see if everything works as expected (I will work on that tomorrow)
- What should happen if the parameters dictionary can't be serialized to a valid JSON? Simply return and don't update the headerData or should there be thrown an error?

Tests/JWSTests.swift
```diff
@@ -54,13 +65,17 @@ class JWSTests: CryptoTestCase {
     // MARK: - RSA Tests
 
     @available(*, deprecated)
-    private func performTestRSASign(algorithm: SignatureAlgorithm, compactSerializedJWS: String) {
```
fbernutz: I decided to delete this argument `compactSerializedJWS: String` because it was not used in the function at all.

JOSESwift/Sources/JWEHeader.swift
daniel-mohemian: > What should happen if it is somehow not a valid JSON?

Good point. We should never get to this point (as `parameters` is internal and your setters guarantee either `String` or `[String]`). The `headerData` is only used for serializing, and most importantly verifying a JWS / decrypting a JWE. We need to keep this `headerData` separate from the parameters dictionary because a Swift dictionary does not guarantee fixed ordering of its values. However, for e.g. signature validation it is crucial that we verify against the header parameters in the order we received them.

I admit, our current implementation is a bit awkward. But for now, we should be fine. As long as `JSONSerialization.isValidJSONObject(parameters)` is true (which is currently guaranteed by your implementation) we will not get in an inconsistent sate where `headerData` and `parameter` differ. We hope to refactor those things in the future.

daniel-mohemian:
We just released [1.3.0](https://github.com/airsidemobile/JOSESwift/releases/tag/1.3.0) which includes your contribution. 🎉 

Make sure to run `pod repo update` if you use JOSESwift via CocoaPods to get the latest changes.

fbernutz:
Perfect! Thanks for merging 🙌

Issue #57, opened 2018-02-20
Research and plan JWK x5 u/c/t parameters


carol-mohemian:

daniel-mohemian:
Check the status of #86 when implementing this.

PR #116, opened 2018-10-12
[86] Ignore `[String]` JWK parameters

### LINKED ISSUE #86: JWK to RSAPublicKey ignore additional parameters
I get JWK from Auth0 JWKSet and transform it to Data object, but when I try to get RSAPublicKey with the data, it throws the following error: 

Thread 1: Fatal error: 'try!' expression unexpectedly raised an error: Swift.DecodingError.typeMismatch(Swift.String, Swift.DecodingError.Context(codingPath: [JWKParameter(stringValue: "x5c", intValue: nil)], debugDescription: "Expected to decode String but found an array instead.", underlyingError: nil))

The problem appears to be "x5c" parameter which has its value in brackets ("x5c": [...]). Since I am getting keys directly from endpoint, the only way to solve this seems to be manually removing the brackets, but since I don't really need "x5c" parameter right now it makes me wonder if there is a way to ignore some parameters?

xaverlohmueller:
## Description

(Fixes #86)

This pull request provides a workaround for the `x5c` parameter, which isn't a `String` but an Array of Strings (`[String]`).
The current behavior is to throw an error when serializing.

Please note that this is just a workaround for #86 and does not do any certificate validation (as planned in #57).


JOSESwift/Sources/RSAKeyCodable.swift
```diff
@@ -62,7 +62,7 @@ extension RSAPublicKey: Decodable {
 
         // Other common parameters are optional.
         var parameters: [String: String] = [:]
-        for key in commonParameters.allKeys {
+        for key in commonParameters.allKeys where key != .X509CertificateChain {
```
daniel-mohemian: Please also ignore the `keyOperations` parameter, [it is an array of strings](https://tools.ietf.org/html/rfc7517#section-4.3) as well.

```suggestion
        for key in commonParameters.allKeys where key != .X509CertificateChain && key != .keyOperations {
```

Tests/JWKx5cTests.swift
```diff
@@ -0,0 +1,37 @@
...
+//  Tests
+//
+//  Created by Xaver Lohmüller on 12.10.18.
+//
+
+import XCTest
+@testable import JOSESwift
+
+class JWKx5cTests: XCTestCase {
+    func testBuildingJWKSetShouldNotFailIfCertificatesArePresent() {
```
daniel-mohemian: This test would probably fit nicely into [`JWKRSADecodingTests.swift`](https://github.com/xaverlohmueller/JOSESwift/blob/96fe42f8a12ed790b2637c870d5671b602a07af4/Tests/JWKRSADecodingTests.swift#L27).

JOSESwift/Sources/RSAKeyCodable.swift
daniel-mohemian: It would be awesome if you could also add the workaround to the decoding of `RSAPrivateKey` [here](https://github.com/xaverlohmueller/JOSESwift/blob/96fe42f8a12ed790b2637c870d5671b602a07af4/JOSESwift/Sources/RSAKeyCodable.swift#L127).

daniel-mohemian: It would be awesome if you could also ignore the `keyOperations` parameter here and in the private key decoding.

```suggestion
    for key in commonParameters.allKeys where key != .X509CertificateChain && key != .keyOperations {
```

JOSESwift/Sources/RSAKeyCodable.swift
```diff
@@ -119,7 +119,7 @@ extension RSAPrivateKey: Decodable {
...
```
daniel-mohemian: It would be awesome if you could also ignore the keyOperations parameter here and in the public key decoding.

```suggestion
        for key in commonParameters.allKeys where key != .X509CertificateChain && key != .keyOperations {
```

JOSESwift/Sources/RSAKeyCodable.swift
xaverlohmueller: Oh, I'm sorry, I must've only read the second half of your comment.

xaverlohmueller:
Awesome!

carol-mohemian:
Thanks for your contribution @xaverlohmueller from my side as well! 👏 

Just FYI: A new version with the merged changes will be released during the day.

daniel-mohemian:
The [1.3.1](https://github.com/airsidemobile/JOSESwift/releases/tag/1.3.1) release is now live on CocoaPods! 🎉

Issue #117, opened 2018-10-17
Allow decoding of `[String]` JWK parameters


daniel-mohemian:
Currently, we ignore (via #116) the JWK parameters `X509CertificateChain` and `keyOperations` because their type is `[String]`. This is because we only decode into a `[String: String]` dictionary at the moment. We should enable the decoding of `String` and `[String]` to be standard compliant. The relevant code is in [`RSAKeysCodable.swift`](https://github.com/airsidemobile/JOSESwift/blob/b6094a35f47faca02542192d5b841f87f046aa95/JOSESwift/Sources/RSAKeyCodable.swift#L66) and in the JWK types themselves e.g. in [`RSAPublicKey.swift`](https://github.com/airsidemobile/JOSESwift/blob/b6094a35f47faca02542192d5b841f87f046aa95/JOSESwift/Sources/RSAKeys.swift#L97).

Possible parameters and their types are listed in [RFC-7517](https://tools.ietf.org/html/rfc7517#section-4):

- "kty": `String`
- "use": `String`
- "key_ops": `[String]`
- "alg": `String`
- "kid": `String`
- "x5u": `URI`
- "x5c": `[String]`
- "x5t": `String`
- "x5t#S256": `String`

PR #118, opened 2018-10-17
Mention ignored JWK parameters in readme

daniel-mohemian:
📃 Mentions that we now ignore the `[String]` JWS parameters that we currently cannot decode (as reported in #86 and tracked to fix in #117).

Once merged, I'll add that note to the wiki as well.


daniel-mohemian:
Ready now @airsidemobile/ios 📄 🚀

PR #124, opened 2018-11-06
Update to Swift 4.2 in more places

xaverlohmueller:
## Description

Deep down in my heart, I knew that pressing a convenient button in Xcode (like I did in #121) wouldn't be enough.

I forgot to bump the Swift version in the `.podspec` file (and maybe other places).

This pull request updates the Swift version in all places, removing the migration hint for good :+1:.


daniel-mohemian:
Awesome thanks a lot Xaver! This is my fault too, I should've seen that before merging. 😅 

I'll merge your pull request not to master directly but to a swift-4-2 feature branch. This is because the Swift 4.2 change apparently breaks the CommonCrypto import. That's not a big issue, though since [it seems like](https://twitter.com/mxcl/status/1003789764884140038) Swift now has a native CommonCrypto module.

I'll take care of throwing out our old CommonCrypto module map and use the new module as soon as possible. Then we should be good to go to release the final Swift 4.2 version. I'll ping you here as soon as your changes are live.

daniel-mohemian:
Sorry for the delay! Your changes are now live under 1.4.0 via #127. Thanks again!

PR #126, opened 2018-11-28
Use native common crypto module

daniel-mohemian:
Swift 4.2 introduced a native common crypto module which we can/must use now. We can't use both or we will get build errors. Apparently, there is no official documentation though. See [here](https://twitter.com/mxcl/status/1003789764884140038).

PR #127, opened 2018-12-04
Update to Swift 4.2 for real

daniel-mohemian:
With the help of @xavierLowmiller we should now be ready for Swift 4.2. Not much has changed, though:

- Update Swift version to 4.2 (via #121 and #124)
- Do automated Xcode project file conversions (via #121 and #124)
- Use native Common crypto module (via #126)


daniel-mohemian:
Thanks guys!

PR #130, opened 2018-12-04
Set new minimum fastlane version to 2.104.0 to fix #129

### LINKED ISSUE #129: Security CVE CVE-2018-1000544
https://nvd.nist.gov/vuln/detail/CVE-2018-1000544

alex-mohemian:
Fixes CVE 2018-1000544 and resolves https://github.com/airsidemobile/JOSESwift/issues/129.

Fastlane `2.104.0` is the first version that has version `1.2.2` of `rubyzip` as a minimum requirement.


daniel-mohemian:
I think the build issue was that Travis doesn't run `bundle update` on every build, so I had to do that locally. We should probably update our `.travis.yml` and add that call.

Anyway, thanks for taking care of the issue @alex-mohemian! 👍

alex-mohemian:
@daniel-mohemian yeah I saw that. But I also wanted to downgrade to the _minimum_ version of fastlane that includes the fix.

Travis doesn't need to do `bundle update`, I should have done that locally, of course but I submitted that from my phone, so... 😉

alex-mohemian:
Since https://github.com/airsidemobile/JOSESwift/pull/132 fixes this, I'm closing the PR.

PR #134, opened 2018-12-24
[107] Add JWE "zip" support

### LINKED ISSUE #107: JWE "zip" support
Are we going to handle compression?
specs: https://tools.ietf.org/html/rfc7516#section-4.1.3

I'm afraid this would involve the addition of dependency to external library like [DataCompression](https://github.com/mw99/DataCompression).

fhaeser:
Resolves #107 

Added support for the 🗜[Compression Algorithm](https://tools.ietf.org/html/rfc7516#section-4.1.3) "zip" header parameter including its handling within the processing chain ⛓.

Due to the fact that there is already a strong dependency on native iOS libraries e.g. `CommonCrypto` I decided to go for the iOS [compression framework](https://developer.apple.com/documentation/compression).

The library [DataCompression](https://github.com/mw99/DataCompression) that @lukaszgieraltowski mentioned in his comment provides a neat wrapper around that framework.
Since there is only one compression algorithm officially supported so far, instead of using the whole library, I went for an excerpt that covers the bare minimum.

The `DeflateCompressor` protocol should provide enough abstraction to exchange it with another implementation if needed.

Notes for the reviewer 🔍 

Compression is performed on the plaintext before encryption and after decryption.
All the relevant base64 de- and encoding is already properly handled.

As the `deprecated` decryption function was not abstracted in the first place, I also did not follow the `DRY` principle.

`Inflate` and `deflate` are force unwrapped as the implementor has to deal with the failing of either of both functions anyhow.

Happy to receive and incorporate suggestions and feedback. 🤓

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,10 @@ merge to master and move them to a respective version upon release.
 
 Add your changes here.
 
+## [1.6.0] - 2019-01-25
+
+- Adds support for RSA-OAEP-256 ([#135](https://github.com/airsidemobile/JOSESwift/pull/135))
+
 ## [1.5.0] - 2019-01-23
 
 - Changes the way elliptic curve keys are decoded to work around [#86](https://github.com/airsidemobile/JOSESwift/issues/86) until [#120](https://github.com/airsidemobile/JOSESwift/pull/120) is merged ([#137](https://github.com/airsidemobile/JOSESwift/pull/137))
--- a/JOSESwift.podspec
+++ b/JOSESwift.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name              = "JOSESwift"
-  s.version           = "1.5.0"
+  s.version           = "1.6.0"
   s.license           = "Apache License, Version 2.0"
   s.summary           = "JOSE framework for Swift"
   s.authors           = { "Daniel Egger" => "daniel.egger@airsidemobile.com", "Carol Capek" => "carol.capek@airsidemobile.com", "Christoph Gigi Fuchs" => "christoph.fuchs@airsidemobile.com" }
--- a/JOSESwift/Sources/CryptoImplementation/RSA.swift
+++ b/JOSESwift/Sources/CryptoImplementation/RSA.swift
@@ -60,22 +60,22 @@ internal extension AsymmetricKeyAlgorithm {
         }
     }
 
-    /**
-     This method returns the maximum message length allow for an Asymmetric Algorithm
-     - parameters:
-        - publicKey: SecKey the publicKey used to acquire the SecKey block size
-     - returns: Int the maximum length allowed
-     discussion:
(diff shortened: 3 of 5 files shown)


dnlggr:
Sweet! That‘s a nice christmas present. 🎄🎁

I‘ll definitely have a look over the next days. 😍

JOSESwift/Sources/JWEHeader.swift
```diff
@@ -211,4 +211,14 @@ extension JWEHeader: CommonHeaderParameterSpace {
             return parameters["crit"] as? [String]
         }
     }
+
+    /// The zip header parameter indicates the content has been compressed before encryption
+    public var zip: String? {
```
dnlggr: Do you think it would make sense to return a `CompressionAlgorithm` instead of a `String` here? It would enable easier switching on the value for the user (without having to convert the String to an enum case or doing string comparison).

JOSESwift/Sources/JWE.swift
```diff
@@ -72,10 +72,19 @@ public struct JWE {
     /// - Throws: `JOSESwiftError` if any error occurs while encrypting.
     public init<KeyType>(header: JWEHeader, payload: Payload, encrypter: Encrypter<KeyType>) throws {
         self.header = header
+        if header.zip != nil && header.zip != CompressionAlgorithm.deflate.rawValue {
```
dnlggr: In my opinion, here and below it would be handy to be able to switch on `zip` instead of doing string comparisons.

JOSESwift/Sources/JWE.swift
```diff
@@ -72,10 +72,19 @@ public struct JWE {
...
+            throw JOSESwiftError.decryptingFailed(description: "Only DEF supported as zip parameter.")
+        }
+        
+        var potentiallyCompressedPayload: Payload = payload
+        // If "zip" parameter is present, compression is applied to the plaintext before encryption.
+        if header.zip == CompressionAlgorithm.deflate.rawValue {
+            potentiallyCompressedPayload = Payload(payload.data().deflate()!)
```
dnlggr: This fore unwrap is not safe. Please throw instead.

JOSESwift/Sources/DeflateCompressor.swift
```diff
@@ -0,0 +1,13 @@
+//
+//  DeflateCompressor.swift
+//  JOSESwift
+//
+//  Created by Florian Häser on 24.12.18.
+//
+
+import Foundation
+
+internal protocol DeflateCompressor {
```
dnlggr: Maybe I'm missing something. Does this protocol do anything / is it used somewhere where it brings abstraction?

daniel-mohemian:
@haeser Did you have a chance yet to look at the feedback?

fhaeser:
Yes thank you for you feedback. But since I am contributing this in my free time I only have the weekends and evenings to incorporate it. Will do it 🔜

fhaeser:
sorry my setup was way to messy will create a new PR incorporating your feedback

JOSESwift/Sources/DeflateCompressor.swift
fhaeser: No you are right. It doesn't bring any benefit. Will remove it.

PR #151, opened 2019-03-17
Add Swiftlint build phase and fix most linting errors

xavierLowmiller:
Hey, it's been a while, so I thought I start tackling #98.

So far, this pull request
* runs SwiftLint autocorrection
* adds a SwiftLint build phase to trigger warnings in Xcode
* fixes/disables the errors that prevent the project from building
* removes the `ignore_exit_status` from the fastlane action

Since this is already getting kind of big, I think it's better to fix the remaining 254 warnings bit by bit.

In order to make this as smooth as possible, we should do one of the following:
1. Disable the offending rules, fix and re-enable them later
1. Remove the SwiftLint build phase, fix the warnings and add it again
1. Disable the biggest offenders (`xctfail_message` and `line_length`), fix them later, and fix the remaining 30 issues now

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -13,17 +13,8 @@ disabled_rules:
   - function_body_length
   - type_body_length
   - file_length
-identifier_name:
-  validates_start_with_lowercase: false
-  excluded:
-    - to
-    - a
-    - b
-    - d
-    - k
-    - n
-    - s
-    - x
-    - y
+  - type_name
+  - identifier_name
+
 line_length:
   ignores_comments: true
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -4,3 +4,21 @@ custom_rules:
     regex: '(\/\*\*)'
     message: "Use `///` instead of `/** ... */` for documentation comments."
     severity: warning
+
+disabled_rules:
+  - force_try
+  - force_cast
+identifier_name:
+  validates_start_with_lowercase: false
+  excluded:
+    - to
+    - a
+    - b
+    - d
+    - k
+    - n
+    - s
+    - x
+    - y
+line_length:
+  ignores_comments: true
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -15,6 +15,7 @@ disabled_rules:
   - file_length
   - type_name
   - identifier_name
+  - todo
 
 line_length:
   ignores_comments: true
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -8,6 +8,11 @@ custom_rules:
 disabled_rules:
   - force_try
   - force_cast
+  - xctfail_message
+  - line_length
+  - function_body_length
+  - type_body_length
+  - file_length
 identifier_name:
   validates_start_with_lowercase: false
   excluded:
--- a/JOSESwift.xcodeproj/project.pbxproj
+++ b/JOSESwift.xcodeproj/project.pbxproj
@@ -546,6 +546,7 @@
 				65FBFDDE1F45CC7C005C7D68 /* Frameworks */,
 				65FBFDDF1F45CC7C005C7D68 /* Headers */,
 				65FBFDE01F45CC7C005C7D68 /* Resources */,
+				4E06574B223CC78D00D7A199 /* SwiftLint */,
 			);
 			buildRules = (
 			);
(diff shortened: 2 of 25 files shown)


daniel-mohemian:
Wow, that's cool! Much appreciated! 😍 After fixing old linting errors in the codebase we can finally enable a strict danger linting setup for new pull requests.

I'll take a look later and let you know if I have any feedback to give. Thanks again in advance!

JOSESwift.xcodeproj/project.pbxproj
```diff
@@ -616,6 +617,27 @@
...
+			inputPaths = (
+			);
+			name = SwiftLint;
+			outputFileListPaths = (
+			);
+			outputPaths = (
+			);
+			runOnlyForDeploymentPostprocessing = 0;
+			shellPath = /bin/sh;
+			shellScript = "if which swiftlint >/dev/null; then\n    swiftlint autocorrect\n    swiftlint\nelse\n    echo \"warning: SwiftLint not installed, download from https://github.com/realm/SwiftLint\"\nfi\n";
```
daniel-mohemian: I'm not sure if we should force the autocorrect in every build step. What would you think of a more passive approach:

- Only linting in the build phase.
- Autocorrect as a Fastlane lane. If we enable danger later, we can let danger remind the user to run the autocorrect lane in case of any linting violations.

My fear with having autocorrect in the build step is that it may also touch things outside the current scope which might make modular commits etc. harder. Let me know what you think and of course, if you disagree!

---

For the warning message, maybe we could add "or run `brew bundle`", since we have `swiftlint` in our `Brewfile`. 👌 🤖

fastlane/Fastfile
```diff
@@ -9,8 +9,7 @@ end
 
 desc "Lint Swift files"
 lane :lint do
-  # Todo: Don't ignore exit status (#98)
-  swiftlint(ignore_exit_status: true)
+  swiftlint()
```
daniel-mohemian:
```suggestion
  swiftlint
```

JOSESwift.xcodeproj/project.pbxproj
daniel-mohemian: I just remembered, we already have a `format_code` lane, no need to add one. 😅 👌 

We run it on every release. 👍

xavierLowmiller: Hey Daniel, thanks for your feedback 👌

I'll happily address the issues you mentioned.

Regarding the autocorrect on every build: In our company, we do this and basically never run into an issue.
But I can see where you're coming from, and I'll turn it off 👍

.swiftlint.yml
```diff
@@ -4,3 +4,26 @@ custom_rules:
...
+  excluded:
+    - to
+    - a
+    - b
+    - d
+    - k
+    - n
+    - s
+    - x
+    - y
```
xavierLowmiller: @daniel-mohemian What do you think about disabling this rule instead of maintaining a long list of exceptions?

A lot of these variable names (and many others, like `iv`) are standards in cryptography, so I'm not sure this rule is adding any value.

Tests/AESDecrypterTests.swift
```diff
@@ -59,12 +59,37 @@ class AESDecrypterTests: RSACryptoTestCase {
 
     /// Tests the `AES` decryption implementation for AES_256_CBC_HMAC_SHA_512 with the test data provided in the [RFC-7518](https://tools.ietf.org/html/rfc7518#appendix-B.3).
     func testDecryptingA256CBCHS512() {
-        let keyData = "00 01 02 03 04 05 06 07 08 09 0a 0b 0c 0d 0e 0f 10 11 12 13 14 15 16 17 18 19 1a 1b 1c 1d 1e 1f 20 21 22 23 24 25 26 27 28 29 2a 2b 2c 2d 2e 2f 30 31 32 33 34 35 36 37 38 39 3a 3b 3c 3d 3e 3f".hexadecimalToData()
-        let additionalAuthenticatedData = "54 68 65 20 73 65 63 6f 6e 64 20 70 72 69 6e 63 69 70 6c 65 20 6f 66 20 41 75 67 75 73 74 65 20 4b 65 72 63 6b 68 6f 66 66 73".hexadecimalToData()
+        let keyData = """
+        00 01 02 03 04 05 06 07 08 09 0a 0b 0c 0d 0e 0f 10 11 12 13 14 15 16 \
+        17 18 19 1a 1b 1c 1d 1e 1f 20 21 22 23 24 25 26 27 28 29 2a 2b 2c 2d \
+        2e 2f 30 31 32 33 34 35 36 37 38 39 3a 3b 3c 3d 3e 3f
+        """.hexadecimalToData()
```
xavierLowmiller: An alternative to having these inline would be to use [NSDataAsset]. @daniel-mohemian what do you think?


[NSDataAsset]: https://nshipster.com/nsdataasset/

JOSESwift.xcodeproj/project.pbxproj
daniel-mohemian: Thanks Xaver that sounds great! 🎉

.swiftlint.yml
daniel-mohemian: Sure. I'll trust your judgement on this. 👍

Tests/AESDecrypterTests.swift
daniel-mohemian: I think that would be cool. 👍 Probably also a lot of work, would it make sense to do it separately, what do you think?

xavierLowmiller: Yeah, I think so too. Another PR, then. I just wanted to mention it.

xavierLowmiller:
Screenshot 2019-03-19 at 14 35 43

...and we've hit zero warnings 🎉

Tests/RSAEncrypterTests.swift
```diff
@@ -109,7 +109,7 @@ class RSAEncrypterTests: RSACryptoTestCase {
         }
     }
 
-    func testEncryptingWithAliceAndBobKey() {
+    func testEncryptingWithAliceAndBobKey() throws {
```
daniel-mohemian: Making tests throw - why did we not think of that in the first place. 🤦‍♂️ Great addition, thanks! 👍 😄

PR #154, opened 2019-04-30
Update to Swift 5

daniel-mohemian:
- Performed automatic update
- Fixed resulting warnings

More info regarding the `data.withUnsafeBytes` deprecation warning can be found [here]( https://forums.swift.org/t/withunsafebytes-data-api-confusion/22142/6).


JOSESwift/Sources/CryptoImplementation/AES.swift
```diff
@@ -116,32 +121,41 @@ internal struct AES {
...
         let dataLength = data.count
+        let keyLength = key.count
+        let ivLength = initializationVector.count
+
+        guard dataLength > 0, keyLength > 0, ivLength > 0 else {
+            throw AESError.cannotPerformOperationOnEmptyDataBuffer
+        }
 
-        //AES's 128 block size is fix for every key length.
+        // AES's 128 block size is fix for every key length and guaranteed not to be 0.
```
nathan-mohemian:
```suggestion
        // AES's 128 block size is fixed for every key length and guaranteed not to be 0.
```

JOSESwift/Sources/JWEHeader.swift
```diff
@@ -89,7 +89,7 @@ public struct JWEHeader: JOSEHeader {
 // Header parameters that are specific to a JWE Header.
 public extension JWEHeader {
     /// The algorithm used to encrypt or determine the value of the Content Encryption Key.
-    public var algorithm: AsymmetricKeyAlgorithm? {
+    var algorithm: AsymmetricKeyAlgorithm? {
```
nathan-mohemian: Why the visibility changes in this file?

Tests/JWEHeaderTests.swift
```diff
@@ -51,23 +51,23 @@ class JWEHeaderTests: XCTestCase {
 
         XCTAssertEqual(header.parameters["enc"] as? String, SymmetricKeyAlgorithm.A256CBCHS512.rawValue)
         XCTAssertEqual(header.parameters["alg"] as? String, AsymmetricKeyAlgorithm.RSA1_5.rawValue)
-        XCTAssertEqual(header.data(), try! JSONSerialization.data(withJSONObject: parameterDictRSA, options: []))
+        XCTAssertEqual(header.data().count, try! JSONSerialization.data(withJSONObject: parameterDictRSA, options: []).count)
```
nathan-mohemian: Why switch to comparing lengths instead of comparing data directly? This seems like a downgrade in terms of test guarantees.

daniel-mohemian: Sorry, I should have pointed out the reason for this change or moved it to another pr.

Comparing the data is flaky since we cannot guarantee the order of the keys in the (JSON) data. Previously the tests used to fail sometimes if we were unlucky and the keys were in different orders. I think I've come up with a better solution for it now.

JOSESwift/Sources/JWEHeader.swift
daniel-mohemian: Symbols in a public extension are public by default.

daniel-mohemian:
@nathan-mohemian Ready for re-review!

Tests/JWEHeaderTests.swift
nathan-mohemian: Aha, thanks for the explanation. In that case I suppose it's reasonable.

Issue #155, opened 2019-05-22
JWS Signing with Secure Enclave keys


martinmitrevski:
Hello,

I've tried to sign JWS ES256 with a key generated from the Secure Enclave (kSecAttrTokenID = kSecAttrTokenIDSecureEnclave).

The signing fails inside EC.swift -> let status = SecKeyRawVerify(publicKey, .sigRaw, digest, digest.count, signatureBytes, curveType.signatureOctetLength).

I've also tried changing the code a bit, with:
SecKeyCreateSignature(privateKey, .ecdsaSignatureMessageX962SHA256, test as CFData, &error) instead. The signing passes, but the JWS is invalid.

Do you have any suggestions regarding this issue?

Thanks,
Martin

daniel-mohemian:
Hi Martin!

Thanks for reporting this issue. As discussed via email I can also reproduce this behavior.

**Some background**

I looked around and found references to [this issue](https://forums.developer.apple.com/thread/98843) on the Apple developer forums and also [this issue](https://github.com/ethanhuang13/CupertinoJWT/issues/6) on GitHub. From what I could gather this seems like a bug in `SecKeyRawSign` in combination with `.sigRaw` as padding and a key from the secure enclave. Apple seems to recommend `SecKeyCreateSignature` over `SecKeyRawSign` as a workaround for this bug according to [this thread](https://forums.developer.apple.com/thread/98843).

As you said, the problem with `SecKeyCreateSignature` is that it creates an invalid JWS signature. That, of course, doesn't mean that the signature is invalid, it's just not in the representation that the JWS RFC defines. As per [RFC-7515](https://tools.ietf.org/html/rfc7515#appendix-A.3):

> The result of the digital signature is the Elliptic Curve (EC) point (R, S), where R and S are unsigned integers.  (...) The JWS Signature is the value R || S.

If we were to use `SecKeyRawSign` with a padding other than `.sigRaw` or even `SecKeyCreateSignature`, the returned signature should be an ASN.1 object from where we would have to parse out the "raw" signature as defined by RFC-7514. For more background see the mentioned [Apple developer forums thread](https://forums.developer.apple.com/thread/98843), [GitHub issue](https://github.com/ethanhuang13/CupertinoJWT/issues/6), or the [SecPadding docs](https://developer.apple.com/documentation/security/secpadding/1617989-pkcs1sha256).

**Now what can we do to fix this situation?**

[...]

PR #156, opened 2019-06-07
Add support for EC256 keys that are stored inside the Secure Enclave

mschwaig:
This PR changes the EC crypto implemenation to use `SecKeyCreateSignature` and `SecKeyVerfiySignature` for signature creation and verification and adds the required conversions to and from BER encoded ASN.1 format for the signatures.
It also introduces negative tests for EC signature verification, to detect trivial false acceptances for the signature validation. This is important because the tests for the signing also rely on the validation implementation.

See #155 for the discussion which lead to this PR.

Please be thorough with the iOS side of the review, since I do not have that much experience as developer for iOS.

### DIFF
--- a/JOSESwift/Sources/CryptoImplementation/EC.swift
+++ b/JOSESwift/Sources/CryptoImplementation/EC.swift
@@ -98,13 +98,19 @@ public enum ECCurveType: String, Codable {
 fileprivate extension Data {
     func withLengthFixedTo(_ octetLength: Int) -> Data {
         let varLength = self.count
+        if varLength > octetLength + 1 {
+            fatalError("Unable to parse ASN.1 Integer")
+        }
         if varLength == octetLength + 1 {
             return self.dropFirst()
+        }
+        if varLength == octetLength {
+            return self
         }
         if varLength < octetLength {
             return Data.init(count: octetLength - varLength) + self
         }
-        return self
+        fatalError("Unable to parse ASN.1 Integer")
     }
 }
 
@@ -200,7 +206,6 @@ internal struct EC {
     /// - Returns: The signature.
     /// - Throws: `ECError` if any errors occur while signing the input data.
     static func sign(_ signingInput: Data, with privateKey: KeyType, and algorithm: SignatureAlgorithm) throws -> Data {
-        // Sign the input as raw elliptic curve coordinates using a hashing algorithm and a private key.
         guard let curveType = algorithm.curveType else {
             throw ECError.invalidCurveDigestAlgorithm
         }
@@ -214,6 +219,7 @@ internal struct EC {
             throw ECError.signingFailed(description: "Error creating signature. (CFError: \(cfErrorRef!.takeRetainedValue()))")
         }
 
+        // unpack BER encoded ASN.1 format signature to raw format as specified for JWS
         let ecSignatureTLV = [UInt8](signature! as Data)
         do {
             let ecSignature = try ecSignatureTLV.read(.sequence)
@@ -224,7 +230,7 @@ internal struct EC {
 
             return fixlenR + fixlenS
         } catch {
-            throw ECError.signingFailed(description: "could not unwrap ASN1 EC signature")
+            throw ECError.signingFailed(description: "could not unpack ASN.1 EC signature")
         }
     }
 
--- a/JOSESwift/Sources/CryptoImplementation/EC.swift
+++ b/JOSESwift/Sources/CryptoImplementation/EC.swift
@@ -95,6 +95,8 @@ public enum ECCurveType: String, Codable {
     }
 }
 
+// converting integers to and from DER encoded ASN.1 as described here:
+// https://docs.microsoft.com/en-us/windows/desktop/seccertenroll/about-integer
 fileprivate extension Data {
     func withLengthFixedTo(_ octetLength: Int) -> Data {
         let varLength = self.count
@@ -112,26 +114,18 @@ fileprivate extension Data {
         }
         fatalError("Unable to parse ASN.1 Integer")
     }
-}
+
+    func fixedLengthToAsn1() -> Data {
+        assert(self.count > 0)
+        let msb : UInt8 = 0b1000_0000
+        let varlen = self.drop { $0 == 0}
+        if (varlen.first! & msb) == msb {
+            return Data(count: 1) + varlen
+        }
+        return varlen
+    }}
 
 fileprivate extension SignatureAlgorithm {
-    typealias DigestFunction = (
-            UnsafeRawPointer?,
-            UInt32,
-            UnsafeMutablePointer<UInt8>?
-    ) -> UnsafeMutablePointer<UInt8>?
-
-    func createDigest(input: Data) throws -> [UInt8] {
-        guard
-                let computedDigestLength = digestLength,
-                let computedDigestFunction = digestFunction
-                else {
-            throw ECError.invalidCurveDigestAlgorithm
-        }
-        var digest = [UInt8](repeating: 0, count: computedDigestLength)
-        _ = computedDigestFunction(Array(input), UInt32(input.count), &digest)
-        return digest
-    }
 
     var secKeyAlgorithm: SecKeyAlgorithm? {
         switch self {
@@ -145,18 +139,6 @@ fileprivate extension SignatureAlgorithm {
             return nil
         }
     }
-    var digestLength: Int? {
-        switch self {
-        case .ES256:
-            return Int(CC_SHA256_DIGEST_LENGTH)
-        case .ES384:
-            return Int(CC_SHA384_DIGEST_LENGTH)
-        case .ES512:
-            return Int(CC_SHA512_DIGEST_LENGTH)
-        default:
-            return nil
-        }
-    }
 
     var curveType: ECCurveType? {
         switch self {
@@ -166,19 +148,6 @@ fileprivate extension SignatureAlgorithm {
             return ECCurveType.P384
         case .ES512:
             return ECCurveType.P521
-        default:
-            return nil
-        }
-    }
-
-    var digestFunction: DigestFunction? {
-        switch self {
-        case .ES256:
-            return CC_SHA256
-        case .ES384:
-            return CC_SHA384
-        case .ES512:
-            return CC_SHA512
         default:
             return nil
         }
@@ -248,14 +217,23 @@ internal struct EC {
         guard let curveType = algorithm.curveType else {
             throw ECError.invalidCurveDigestAlgorithm
         }
-        let digest = try algorithm.createDigest(input: verifyingInput)
-        let signatureBytes: [UInt8] = Array(signature)
-        let status = SecKeyRawVerify(publicKey, .sigRaw, digest, digest.count, signatureBytes, curveType.signatureOctetLength)
-        if status != errSecSuccess {
-            throw ECError.verifyingFailed(description: "Error validating signature. (OSStatus: \(status))")
+        guard let secKeyAlgorithm = algorithm.secKeyAlgorithm else {
+            throw ECError.algorithmNotSupported
         }
 
-        return true
+        // pack raw signature as apecified for JWS into BER encoded ASN.1 format
+        let fixlenR = signature.prefix(curveType.coordinateOctetLength)
+        let varlenR = [UInt8](fixlenR.fixedLengthToAsn1())
+        let fixlenS = signature.suffix(curveType.coordinateOctetLength)
+        let varlenS = [UInt8](fixlenS.fixedLengthToAsn1())
+        let asn1Signature = Data((varlenR.encode(as: .integer) + varlenS.encode(as: .integer)).encode(as: .sequence))
+
+        var cfErrorRef: Unmanaged<CFError>?
+        let isValid = SecKeyVerifySignature(publicKey, secKeyAlgorithm, verifyingInput as CFData, asn1Signature as CFData, &cfErrorRef)
+        if cfErrorRef != nil {
+            throw ECError.signingFailed(description: "Error verifying signature. (CFError: \(cfErrorRef!.takeRetainedValue()))")
+        }
+        return isValid
     }
 
 }
--- a/JOSESwift/Sources/CryptoImplementation/EC.swift
+++ b/JOSESwift/Sources/CryptoImplementation/EC.swift
@@ -95,6 +95,19 @@ public enum ECCurveType: String, Codable {
     }
 }
 
+fileprivate extension Data {
+    func withLengthFixedTo(_ octetLength: Int) -> Data {
+        let varLength = self.count
+        if varLength == octetLength + 1 {
+            return self.dropFirst()
+        }
+        if varLength < octetLength {
+            return Data.init(count: octetLength - varLength) + self
+        }
+        return self
+    }
+}
+
 fileprivate extension SignatureAlgorithm {
     typealias DigestFunction = (
             UnsafeRawPointer?,
@@ -114,6 +127,18 @@ fileprivate extension SignatureAlgorithm {
         return digest
     }
 
+    var secKeyAlgorithm: SecKeyAlgorithm? {
+        switch self {
+        case .ES256:
+            return .ecdsaSignatureMessageX962SHA256
+        case .ES384:
+            return .ecdsaSignatureMessageX962SHA384
+        case .ES512:
+            return .ecdsaSignatureMessageX962SHA512
+        default:
+            return nil
+        }
+    }
     var digestLength: Int? {
         switch self {
         case .ES256:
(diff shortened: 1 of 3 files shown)


JOSESwift/Sources/CryptoImplementation/EC.swift
```diff
@@ -95,33 +95,46 @@ public enum ECCurveType: String, Codable {
...
+        if varLength < octetLength {
+            return Data.init(count: octetLength - varLength) + self
+        }
+        fatalError("Unable to parse ASN.1 Integer")
     }
 
-    var digestLength: Int? {
+    func fixedLengthToAsn1() -> Data {
+        assert(self.count > 0)
+        let msb : UInt8 = 0b1000_0000
```
Jose-ElRobot: <table data-meta="generated_by_danger">
  <tbody>
    <tr>
      <td>:warning:</td>
      <td width="100%" data-sticky="false"><span data-href="https://github.com/airsidemobile/JOSESwift/blob/98f7c96953234a8dd81f281a2edf38f9ae258880/JOSESwift/Sources/CryptoImplementation/EC.swift#L120"></span>Colons should be next to the identifier when specifying a type and next to the key in dictionary literals.<br />
<code>colon</code> <code>EC.swift:120</code></td>
    </tr>
  </tbody>
</table>

JOSESwift/Sources/CryptoImplementation/EC.swift
```diff
@@ -95,33 +95,46 @@ public enum ECCurveType: String, Codable {
...
+        let varlen = self.drop { $0 == 0}
+        if (varlen.first! & msb) == msb {
```
Jose-ElRobot: <table data-meta="generated_by_danger">
  <tbody>
    <tr>
      <td>:warning:</td>
      <td width="100%" data-sticky="false"><span data-href="https://github.com/airsidemobile/JOSESwift/blob/98f7c96953234a8dd81f281a2edf38f9ae258880/JOSESwift/Sources/CryptoImplementation/EC.swift#L122"></span>Force unwrapping should be avoided.<br />
<code>force_unwrapping</code> <code>EC.swift:122</code></td>
    </tr>
  </tbody>
</table>

JOSESwift/Sources/CryptoImplementation/EC.swift
```diff
@@ -175,25 +175,32 @@ internal struct EC {
...
         }
 
-        let signatureBytes = signature.mutableBytes.assumingMemoryBound(to: UInt8.self)
-        let status = SecKeyRawSign(privateKey, .sigRaw, digest, digest.count, signatureBytes, &signatureLength)
-        if status != errSecSuccess {
-            throw ECError.signingFailed(description: "Error creating signature. (OSStatus: \(status))")
+        var cfErrorRef: Unmanaged<CFError>?
+        let signature = SecKeyCreateSignature(privateKey, secKeyAlgorithm, signingInput as CFData, &cfErrorRef)
+        if cfErrorRef != nil {
+            throw ECError.signingFailed(description: "Error creating signature. (CFError: \(cfErrorRef!.takeRetainedValue()))")
```
Jose-ElRobot: <table data-meta="generated_by_danger">
  <tbody>
    <tr>
      <td>:warning:</td>
      <td width="100%" data-sticky="false"><span data-href="https://github.com/airsidemobile/JOSESwift/blob/98f7c96953234a8dd81f281a2edf38f9ae258880/JOSESwift/Sources/CryptoImplementation/EC.swift#L188"></span>Force unwrapping should be avoided.<br />
<code>force_unwrapping</code> <code>EC.swift:188</code></td>
    </tr>
  </tbody>
</table>

JOSESwift/Sources/CryptoImplementation/EC.swift
```diff
@@ -175,25 +175,32 @@ internal struct EC {
...
         }
 
-        return signature as Data
+        // unpack BER encoded ASN.1 format signature to raw format as specified for JWS
+        let ecSignatureTLV = [UInt8](signature! as Data)
```
Jose-ElRobot: <table data-meta="generated_by_danger">
  <tbody>
    <tr>
      <td>:warning:</td>
      <td width="100%" data-sticky="false"><span data-href="https://github.com/airsidemobile/JOSESwift/blob/98f7c96953234a8dd81f281a2edf38f9ae258880/JOSESwift/Sources/CryptoImplementation/EC.swift#L192"></span>Force unwrapping should be avoided.<br />
<code>force_unwrapping</code> <code>EC.swift:192</code></td>
    </tr>
  </tbody>
</table>

JOSESwift/Sources/CryptoImplementation/EC.swift
```diff
@@ -210,14 +217,23 @@ internal struct EC {
...
+        let fixlenR = signature.prefix(curveType.coordinateOctetLength)
+        let varlenR = [UInt8](fixlenR.fixedLengthToAsn1())
+        let fixlenS = signature.suffix(curveType.coordinateOctetLength)
+        let varlenS = [UInt8](fixlenS.fixedLengthToAsn1())
+        let asn1Signature = Data((varlenR.encode(as: .integer) + varlenS.encode(as: .integer)).encode(as: .sequence))
+
+        var cfErrorRef: Unmanaged<CFError>?
+        let isValid = SecKeyVerifySignature(publicKey, secKeyAlgorithm, verifyingInput as CFData, asn1Signature as CFData, &cfErrorRef)
+        if cfErrorRef != nil {
+            throw ECError.signingFailed(description: "Error verifying signature. (CFError: \(cfErrorRef!.takeRetainedValue()))")
```
Jose-ElRobot: <table data-meta="generated_by_danger">
  <tbody>
    <tr>
      <td>:warning:</td>
      <td width="100%" data-sticky="false"><span data-href="https://github.com/airsidemobile/JOSESwift/blob/98f7c96953234a8dd81f281a2edf38f9ae258880/JOSESwift/Sources/CryptoImplementation/EC.swift#L234"></span>Force unwrapping should be avoided.<br />
<code>force_unwrapping</code> <code>EC.swift:234</code></td>
    </tr>
  </tbody>
</table>

daniel-mohemian:
Awesome, thanks @mschwaig! 🎉 

I'll get back to you with a review early next week.

JOSESwift/Sources/CryptoImplementation/EC.swift
daniel-mohemian: Here you could get rid of the force unwrapping by doing

``` swift
if let error = cfErrorRef {
    throw ... error.takeretainedValue() ...
}

File: JOSESwift/Sources/CryptoImplementation/EC.swift [hunk 6]
```diff
@@ -95,33 +95,46 @@ public enum ECCurveType: String, Codable {
...
-            UnsafeMutablePointer<UInt8>?
-    ) -> UnsafeMutablePointer<UInt8>?
-
-    func createDigest(input: Data) throws -> [UInt8] {
-        guard
-                let computedDigestLength = digestLength,
-                let computedDigestFunction = digestFunction
-                else {
-            throw ECError.invalidCurveDigestAlgorithm
+// converting integers to and from DER encoded ASN.1 as described here:
```
daniel-mohemian: It would be great if you could add a short explanation of why exactly this padding is needed. Similar to your explanation for https://github.com/ethanhuang13/CupertinoJWT/issues/12 would be great. 👍

JOSESwift/Sources/CryptoImplementation/EC.swift
```diff
@@ -95,33 +95,46 @@ public enum ECCurveType: String, Codable {
...
+            return Data.init(count: octetLength - varLength) + self
+        }
+        fatalError("Unable to parse ASN.1 Integer")
     }
 
-    var digestLength: Int? {
+    func fixedLengthToAsn1() -> Data {
+        assert(self.count > 0)
+        let msb : UInt8 = 0b1000_0000
+        let varlen = self.drop { $0 == 0}
```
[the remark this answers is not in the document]
mschwaig: This should just drop leading zeros so that's a bug.

JOSESwift/Sources/CryptoImplementation/EC.swift
```diff
@@ -175,25 +175,32 @@ internal struct EC {
...
+        // unpack BER encoded ASN.1 format signature to raw format as specified for JWS
+        let ecSignatureTLV = [UInt8](signature! as Data)
+        do {
+            let ecSignature = try ecSignatureTLV.read(.sequence)
+            let varlenR = try Data(ecSignature.read(.integer))
+            let varlenS = try Data(ecSignature.skip(.integer).read(.integer))
+            let fixlenR = varlenR.withLengthFixedTo(curveType.coordinateOctetLength)
+            let fixlenS = varlenS.withLengthFixedTo(curveType.coordinateOctetLength)
+
+            return fixlenR + fixlenS
```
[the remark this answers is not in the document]
mschwaig: > One more suggestion from my side: What would you think about moving the DER/ASN.1 packing and unpacking code to functions that can be tested independently from the signing/verifying code
> logic? Since this is the crucial feature of this pull request, it would be awesome to be able to test them independently.

I think testing this logic independently would be a good idea, but I would not want to extend the implementation to support parsing more than what is strictly required for this use case or have these functions exposed to other parts of the code in a way that suggests they can be reused elsewhere when actually right now they only work within the constraints that apply for this use case.

I have to look into how to scope things in Swift to achieve this and have it be testable.

JOSESwift/Sources/CryptoImplementation/EC.swift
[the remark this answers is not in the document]
mschwaig: Checking the API docs I think this is fine: https://developer.apple.com/documentation/foundation/data/2832697-drop

That case can also be covered with a test if I manage to write tests for those conversions.

JOSESwift/Sources/CryptoImplementation/EC.swift
mschwaig: Should I opt to cast and throw the retained value or add it to the description string like I am currently doing?

JOSESwift/Sources/CryptoImplementation/EC.swift
mschwaig: Is there a preferable alternative to the force unwrap here?
The signature should always be non-null at this point since the contract of `SecKeyCreateSignature` only allow `nil` here on failure.
If not I would add a comment and disable the linter here.

JOSESwift/Sources/CryptoImplementation/EC.swift
```diff
@@ -95,33 +95,46 @@ public enum ECCurveType: String, Codable {
...
+            return self.dropFirst()
         }
-        var digest = [UInt8](repeating: 0, count: computedDigestLength)
-        _ = computedDigestFunction(Array(input), UInt32(input.count), &digest)
-        return digest
+        if varLength == octetLength {
+            return self
+        }
+        if varLength < octetLength {
+            return Data.init(count: octetLength - varLength) + self
```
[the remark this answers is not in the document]
mschwaig: About the documentation of the padding:

As far as I can tell the [JWS standard only says](https://tools.ietf.org/html/rfc7515#appendix-A.3.1):
> the R and S values, given as octet sequences representing big-endian integers

and later

> The JWS Signature is the value R || S

R and S need to have a fixed length, because the length of R and S is not encoded anywhere else explicitly so otherwise there is no way for the verifying party to tell where R ends and S begins.

I cannot find this mentioned more explicitly in the standard.

JOSESwift/Sources/CryptoImplementation/EC.swift
```diff
@@ -95,33 +95,46 @@ public enum ECCurveType: String, Codable {
...
+        }
+        fatalError("Unable to parse ASN.1 Integer")
     }
 
-    var digestLength: Int? {
+    func fixedLengthToAsn1() -> Data {
+        assert(self.count > 0)
```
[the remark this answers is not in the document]
mschwaig: Yes, asserts are omitted in release builds.

Instead of changing the assert, I will add an additional check directly in the verify method, so incorrect user code gets an error and the assert is only an internal thing.

JOSESwift/Sources/CryptoImplementation/EC.swift
```diff
@@ -175,25 +187,32 @@ internal struct EC {
...
-            throw ECError.signingFailed(description: "Error creating signature. (OSStatus: \(status))")
+        var cfErrorRef: Unmanaged<CFError>?
+        let signature = SecKeyCreateSignature(privateKey, secKeyAlgorithm, signingInput as CFData, &cfErrorRef)
+        if let error = cfErrorRef {
+            throw ECError.signingFailed(description: "Error creating signature. (CFError: \(error.takeRetainedValue()))")
         }
 
-        return signature as Data
+        // unpack BER encoded ASN.1 format signature to raw format as specified for JWS
+        let ecSignatureTLV = [UInt8](signature! as Data)
```
Jose-ElRobot: <table data-meta="generated_by_danger">
  <tbody>
    <tr>
      <td>:warning:</td>
      <td width="100%" data-sticky="false"><span data-href="https://github.com/airsidemobile/JOSESwift/blob/d5682dbc1f03c0452bb5d60e2fbe2dda54b43725/JOSESwift/Sources/CryptoImplementation/EC.swift#L204"></span>Force unwrapping should be avoided.<br />
<code>force_unwrapping</code> <code>EC.swift:204</code></td>
    </tr>
  </tbody>
</table>

JOSESwift/Sources/CryptoImplementation/EC.swift
```diff
@@ -206,18 +225,30 @@ internal struct EC {
...
         }
-        let digest = try algorithm.createDigest(input: verifyingInput)
-        let signatureBytes: [UInt8] = Array(signature)
-        let status = SecKeyRawVerify(publicKey, .sigRaw, digest, digest.count, signatureBytes, curveType.signatureOctetLength)
-        if status != errSecSuccess {
-            throw ECError.verifyingFailed(description: "Error validating signature. (OSStatus: \(status))")
+        guard let secKeyAlgorithm = algorithm.secKeyAlgorithm else {
+            throw ECError.algorithmNotSupported
+        }
+        if (signature.count != curveType.coordinateOctetLength * 2) {
```
Jose-ElRobot: <table data-meta="generated_by_danger">
  <tbody>
    <tr>
      <td>:warning:</td>
      <td width="100%" data-sticky="false"><span data-href="https://github.com/airsidemobile/JOSESwift/blob/d5682dbc1f03c0452bb5d60e2fbe2dda54b43725/JOSESwift/Sources/CryptoImplementation/EC.swift#L235"></span><code>if</code>, <code>for</code>, <code>guard</code>, <code>switch</code>, <code>while</code>, and <code>catch</code> statements shouldn’t unnecessarily wrap their conditionals or arguments in parentheses.<br />
<code>control_statement</code> <code>EC.swift:235</code></td>
    </tr>
  </tbody>
</table>

JOSESwift/Sources/CryptoImplementation/EC.swift
```diff
@@ -95,33 +95,46 @@ public enum ECCurveType: String, Codable {
...
-                let computedDigestFunction = digestFunction
-                else {
-            throw ECError.invalidCurveDigestAlgorithm
+// converting integers to and from DER encoded ASN.1 as described here:
+// https://docs.microsoft.com/en-us/windows/desktop/seccertenroll/about-integer
+fileprivate extension Data {
+    func withLengthFixedTo(_ octetLength: Int) -> Data {
+        let varLength = self.count
+        if varLength > octetLength + 1 {
+            fatalError("Unable to parse ASN.1 Integer")
```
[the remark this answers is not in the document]
mschwaig: This will only happen Apple violates the contract of their `SecKeyCreateSignature` function (or we made an implementation error), user code cannot trigger this condition.

How do you handle such cases elsewhere?

JOSESwift/Sources/CryptoImplementation/EC.swift
mschwaig: Done.

JOSESwift/Sources/CryptoImplementation/EC.swift
```diff
@@ -95,33 +95,46 @@ public enum ECCurveType: String, Codable {
...
+        }
+        if varLength == octetLength + 1 {
```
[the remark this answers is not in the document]
mschwaig: I added an assert which checks this.

JOSESwift/Sources/CryptoImplementation/EC.swift
mschwaig: This was a relevant edge case for which I added a code path.

JOSESwift/Sources/CryptoImplementation/EC.swift
mschwaig: That's great, thanks. I did that.

JOSESwift/Sources/CryptoImplementation/EC.swift
mschwaig: I did the description string thing for now.

JOSESwift/Sources/CryptoImplementation/EC.swift
mschwaig: Found another way to fix address this with an `if let` inside a `guard let` and a `fatalError` if they mismatch.

JOSESwift/Sources/CryptoImplementation/EC.swift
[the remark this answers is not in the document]
daniel-mohemian: That is a good point. For testing, we would have to raise the access control level from `fileprivate` to `internal`. I totally agree with you that that might suggest that the functions could be used from elsewhere in the code base- That's certainly not what we want.

What would you think of having them as an extension of `ECCurveType` that would clearly namespace it as something `EC` internal and would even give you direct access to the `coordinateOctetLength`. Just an idea. Let me know what you think, I'm open for suggestions.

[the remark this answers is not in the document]
mschwaig: I am halfway-done with the tests on my work machine. Currently I am using another struct inside of the `EC` struct with two static functions to avoid the extension getting exposed everywhere, what you are suggesting sounds like it may be a better idea.

mschwaig:
> Great job @mschwaig, thanks for going over it again and resolving all remaining issues.
> 
> Let me know what you think regarding testing the two asn1/fixedlength functions. If you want we can also track an issue for that and add it in a separate pull request. If you want I can handle that for you.

I am about halfway done with those tests. I prepared test data for all of the cases I think are relevant, but I did not implement any actual test cases yet.
I will only be able to continue working on them after NovaRock. Adding the tests in a seperate PR would be good I think, but either way is fine for me. Feel free to proceed with this however you like.

mschwaig:
One thing I am still concerned about is if the EC implementation still behaves in the same way as before for negative cases like invalid signatures or blatantly incorrectly formatted signatures, since the verify method can return false but also throw errors. This negative case behaviour might be inconsistent between the old and new implementation. I cannot judge if this would be an issue, I remember thinking maybe also differs for RSA and EC right now, but I did not investigate further.

JOSESwift/Sources/CryptoImplementation/EC.swift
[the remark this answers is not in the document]
daniel-mohemian: I'll leave it up to you how to best handle this. 👍

daniel-mohemian:
> One thing I am still concerned about is if the EC implementation still behaves in the same way as before for negative cases like invalid signatures or blatantly incorrectly formatted signatures, since the verify method can return false but also throw errors.

Good point. Your implementation now behaves as its RSA counterpart. It returns true/false for valid/invalid signatures and throws in case of an error. The old `SecKeyRawVerify` did not return a boolean but only a status code. I think mirroring the behavior of the newer `SecKeyVerifySignature`, as you did, is the way to go. What do you think?

I will only be able to continue working on them after NovaRock. Adding the tests in a seperate PR would be good I think, but either way is fine for me. Feel free to proceed with this however you like.

Ok, cool! I'll merge this pull request now but hold off on releasing the change until the tests are in. That way we get smaller and easier reviewable pull requests while still maintaining the quality of the release. 

Have fun at Nova Rock! ☀️ 🇦🇹 🎸 🍻

mschwaig:
Sounds great, thanks. 😁

daniel-mohemian:
For completeness, I tracked #157. Let me know if you need anything!

PR #159, opened 2019-06-17
Add security policy

daniel-mohemian:
It is now recommended by Github :octocat: to have a security policy.

This is basically what we had in the readme so far. I just moved it so everything is in one place and rephrased some sentences to be more clear.

We also had an error in the algorithm table and forgot to add a ✅ once. Fixed that as well.


SECURITY.md
```diff
@@ -0,0 +1,17 @@
...
+
+However, we might consider extending the library in a way that would allow the current cryptographic implementation (based on Apple’s SDKs) to _optionally_ be switched out by a different implementation if this enables JOSESwift to run in a pure Swift environment. Please contact us if this is something you are thinking about.
+
+## Supported Versions
+
+Please make sure to always update to the latest version to receive security related patches.
+
+## Reporting a Vulnerability
+
+For security disclosures or related matters, please contact <joseswift@airsidemobile.com>. We will do our best to review and respond to security disclosures with the highest priority.
```
equalspeterd: Is it worth indicating security scanning practices (SonarQube)?

daniel-mohemian: Sure, good point. How would you phrase it?

Fyi: In addition to scanning on Airside's private Sonarqube server, we're also planning to set up automated and transparent scanning of JOSESwift using Sonarcloud. But until that is set up let's add a sentence or two about the private Sonarqube scanning to this document.

PR #186, opened 2019-10-31
Add dependency checker

carol-mohemian:
I figured we didn't add the dependency checker here. I already had a brief discussion with @daniel-mohemian because GitHub's vulnerability check is already quite good but I think it doesn't support CocoaPods. I will check and also if everything works. Daniel reported that Travis CI is somehow broke. Maybe I can fix this in this PR.


daniel-mohemian:
Thanks @carol-mohemian. As discussed, I'll close this for now as we don't have any CocoaPod dependencies.

PR #198, opened 2020-02-05
Dry up signing roundtrip tests

### LINKED ISSUE #189: Fix performTestRSASign function
https://github.com/airsidemobile/JOSESwift/blob/b2e314ace9e9773246b5dbaa18f9e4402bd52b43/Tests/JWSRSATests.swift#L85

We have a pretty serious test bug in the line above where we are not properly checking against the correct expected value. Fix this function so that it gets passed the expected value as parameter.

daniel-mohemian:
Closes #189. We had two helper functions that did basically the same. Dried those up.


Tests/JWSRSATests.swift
```diff
@@ -149,7 +128,19 @@ class JWSRSATests: RSACryptoTestCase {
         let verifier = Verifier(verifyingAlgorithm: algorithm, publicKey: publicKeyAlice2048!)
 
         XCTAssertTrue(secondJWS.isValid(for: verifier!))
-        XCTAssertEqual(String(data: jws.header.data(), encoding: .utf8), "{\"alg\":\"\(algorithm.rawValue)\"}")
+        if withKid {
+            let algKidHeader = "{\"alg\":\"\(algorithm.rawValue)\",\"kid\":\"kid\"}"
+            let kidAlgHeader = "{\"kid\":\"kid\",\"alg\":\"\(algorithm.rawValue)\"}"
```
nathan-mohemian: I think you could make these string literals a little cleaner to read by using [extended string delimiters](https://docs.swift.org/swift-book/LanguageGuide/StringsAndCharacters.html#ID606). Your call whether it's worth the trade off for noisier string interpolation.

Something like 
```swift
let algKidHeader = #"{"alg":"\#(algorithm.rawValue)","kid":"kid"}"#
```

Tests/JWSRSATests.swift
```diff
@@ -149,7 +128,19 @@ class JWSRSATests: RSACryptoTestCase {
         let verifier = Verifier(verifyingAlgorithm: algorithm, publicKey: publicKeyAlice2048!)
 
         XCTAssertTrue(secondJWS.isValid(for: verifier!))
-        XCTAssertEqual(String(data: jws.header.data(), encoding: .utf8), "{\"alg\":\"\(algorithm.rawValue)\"}")
+        if withKid {
```
nathan-mohemian: This if statement wants to be a guard. If you move the check from line 144 up above the branch then an early return wouldn't matter, right?

Tests/JWSRSATests.swift
daniel-mohemian: Agreed, but since we have lots and lots of such strings throughout the project, I tracked https://github.com/airsidemobile/JOSESwift/issues/202.

Tests/JWSRSATests.swift
daniel-mohemian: Updated ✅

daniel-mohemian:
@nathan-mohemian Please re-review!

Tests/JWSRSATests.swift
```diff
@@ -149,8 +128,21 @@ class JWSRSATests: RSACryptoTestCase {
         let verifier = Verifier(verifyingAlgorithm: algorithm, publicKey: publicKeyAlice2048!)
 
         XCTAssertTrue(secondJWS.isValid(for: verifier!))
-        XCTAssertEqual(String(data: jws.header.data(), encoding: .utf8), "{\"alg\":\"\(algorithm.rawValue)\"}")
-        XCTAssertEqual(String(data: jws.payload.data(), encoding: .utf8), "The true sign of intelligence is not knowledge but imagination.")
-    }
+        XCTAssertEqual(String(data: secondJWS.payload.data(), encoding: .utf8), "The true sign of intelligence is not knowledge but imagination.")
```
nathan-mohemian: One tiny thing: was it intentional to switch this check to test `secondJWS` instead of `jws`? I'm assuming this is the "same" line as [153 in the previous version](https://github.com/airsidemobile/JOSESwift/pull/198/files#diff-be48fadbbd6c98cbf736aafe1313bb8bL153).

daniel-mohemian: Yep intentional as it makes a stronger assertion. The second `secondJWS` is the deserialized version of the first `jws`. With that change that we include the deserialization in the round trip.

nathan-mohemian: Ok, looks good to me then 👍

PR #205, opened 2020-02-15
Add CircleCI

haeser:
Replicates our current CI setup in CircleCI. Runs on each PR and the default branch.

Realized as a one by one translation of the GitHub Actions file introduced in https://github.com/airsidemobile/JOSESwift/pull/200

Execution results are accessible for contributors not belonging to the organization.

Screen Shot 2020-02-15 at 16 04 57


.circleci/config.yml
```diff
@@ -0,0 +1,71 @@
...
+            - gem-cache-v1
+      - run:
+          name: Bundler
+          command: |
+            bundle install --path vendor/bundle --jobs 4 --retry 3
+      - save_cache:
+          key: gem-cache-v1-{{ checksum "Gemfile.lock" }}
+          paths:
+            - vendor/bundle
+      - persist_to_workspace:
```
daniel-mohemian: This persists the whole workspace right? I don't think we really need that. Is it possible to not store the whole workspace but rather to restore the cache we save in this step in the other jobs?

.circleci/config.yml
```diff
@@ -0,0 +1,71 @@
+version: 2.1
+
+executors:
+  mac-executor:
+    macos:
+      xcode: "11.2.1"
+    working_directory: ~/code
```
daniel-mohemian:
```suggestion
    working_directory: ~/joseswift
```

.circleci/config.yml
```diff
@@ -0,0 +1,71 @@
...
+            - vendor/bundle
+      - persist_to_workspace:
+          root: .
+          paths:
+            - .
+  test:
+    executor: mac-executor
+    steps:
+      - checkout
+      - attach_workspace:
```
daniel-mohemian: As mentioned, rather than doing this can we just restore the bundler cache here?

.circleci/config.yml
```diff
@@ -0,0 +1,71 @@
...
+            bundle config --local path vendor/bundle
+            bundle exec fastlane sonarqube
+      - store_test_results:
+          path: fastlane/test_output
+
+  danger:
+    executor: mac-executor
+    steps:
+      - checkout
+      - attach_workspace:
```
daniel-mohemian: Same here regarding the cache.

.circleci/config.yml
```diff
@@ -0,0 +1,71 @@
...
+          command: |
+            bundle config --local path vendor/bundle
+            bundle exec fastlane test
+      - run:
+          name: Sonarqube
+          command: |
+            bundle config --local path vendor/bundle
+            bundle exec fastlane sonarqube
+      - store_test_results:
+          path: fastlane/test_output
```
daniel-mohemian: I couldn't find the test summary on CircleCI, maybe we need to rename the ` report.junit` to `report.xml` for it to work? According to the docs, it should work like this, only difference is that [their files have ` .xml` as file endings](https://circleci.com/docs/2.0/configuration-reference/#store_test_results).

haeser: That is because all tests passed. Failing tests would show up here as well. fastlane produces JUnit XML compatible output all good. Documentation can be found [here](https://circleci.com/docs/2.0/collect-test-data/).

Screen Shot 2020-02-17 at 10 11 50

.circleci/config.yml
haeser: Workspaces are another way of caching and the preferred way when passing data between jobs in a workflow [0]. You are right as long as we require data from package dependency managers only, caching is sufficient. Though, as soon as we need e.g. build output for further processing workspaces become a must. An example is displaying tests that have failed right within the PR [1]. You would not cache intermediate outputs using the caching mechanism but pass it in the workspace.

[0] https://circleci.com/docs/2.0/caching/
[1] https://github.com/orta/danger-junit

daniel-mohemian: > An example is displaying tests that have failed right within the PR [1].

Yes but then we would explicitly persist the files we need not the whole repository. I'm not against persisting exactly what we need, but persisting the whole working directory (`.` means we persist _everything_ right?) just because we might need some files in it sometime in the future feels a bit imprecise.

Would the following work:
- Cache and restore dependencies between jobs
- Once we need to persist files between jobs, which are not a fit for the cache use `persist_to_workspace`. An example would be to run Sonarqube in a separate job after the tests and persist just the test report in the work space so that Sonarqube can use it for the coverage calculation

haeser: Yes would absolutely work. Also yes you would only persist the relevant folder/files to the workspace. `.` is indeed too much here. Will do the change.

JOSESwift/Sources/Algorithms.swift
```diff
@@ -40,6 +40,7 @@ public enum SignatureAlgorithm: String {
     case ES512
 }
 
+
```
JOSESwiftCI: <table data-meta="generated_by_danger">
  <tbody>
    <tr>
      <td>:warning:</td>
      <td width="100%" data-sticky="false"><span data-href="https://github.com/airsidemobile/JOSESwift/blob/7e9896f3cc484ba08f327865cabbfb4ea34cddb8/JOSESwift/Sources/Algorithms.swift#L43"></span>Limit vertical whitespace to a single empty line. Currently 2.<br />
<code>vertical_whitespace</code> <code>Algorithms.swift:43</code></td>
    </tr>
  </tbody>
</table>

PR #211, opened 2020-02-18
Add A128KW, A192KW, and A256KW algorithms

daniel-mohemian:
This integrates the changes introduced by #172 int the refactored JWE encryption/decryption flow.

### DIFF
--- a/JOSESwift/Sources/CryptoImplementation/AES.swift
+++ b/JOSESwift/Sources/CryptoImplementation/AES.swift
@@ -36,9 +36,11 @@ fileprivate extension ContentEncryptionAlgorithm {
     var ccAlgorithm: CCAlgorithm {
         switch self {
         case .A256CBCHS512:
+            // 128 referes to the block size in this case.
             return CCAlgorithm(kCCAlgorithmAES128)
 
         case .A128CBCHS256:
+            // 128 referes to the block size in this case.
             return CCAlgorithm(kCCAlgorithmAES128)
         }
     }
@@ -55,27 +57,6 @@ fileprivate extension ContentEncryptionAlgorithm {
 }
 
 fileprivate extension KeyManagementAlgorithm {
-    var ccAlgorithm: CCAlgorithm? {
-        switch self {
-        case .A128KW, .A192KW, .A256KW:
-            return CCAlgorithm(kCCAlgorithmAES)
-        default:
-            return nil
-        }
-    }
-
-    var iv: Data? {
-        switch self {
-        case .A128KW, .A192KW, .A256KW:
-            // See https://tools.ietf.org/html/rfc3394#section-2.2.3.1
-            // The default iv is defined to be the hexadecimal constant A6A6A6A6A6A6A6A6
-            // Todo: Check actual runtime value
-            return Data(bytes: CCrfc3394_iv, count: CCrfc3394_ivLen)
-        default:
-            return nil
-        }
-    }
-
     func checkAESKeyLength(for key: Data) -> Bool? {
         switch self {
         case .A128KW:
@@ -114,7 +95,7 @@ enum AES {
                 throw AESError.keyLengthNotSatisfied
             }
 
-            let encrypted = try ccAESCBCCrypt(
+            let encrypted = ccAESCBCCrypt(
                 operation: CCOperation(kCCEncrypt),
                 data: plaintext, key: encryptionKey,
                 algorithm: algorithm.ccAlgorithm,
@@ -122,11 +103,14 @@ enum AES {
                 padding: CCOptions(kCCOptionPKCS7Padding)
             )
 
-            guard encrypted.status == UInt32(kCCSuccess) else {
+            guard
+                let ciphertext = encrypted.data,
+                encrypted.status == UInt32(kCCSuccess)
+            else {
                 throw AESError.encryptingFailed(description: "Encryption failed with status: \(encrypted.status).")
             }
 
-            return encrypted.data
+            return ciphertext
         }
     }
 
@@ -151,7 +135,7 @@ enum AES {
                 throw AESError.keyLengthNotSatisfied
             }
 
-            let decrypted = try ccAESCBCCrypt(
+            let decrypted = ccAESCBCCrypt(
                 operation: CCOperation(kCCDecrypt),
                 data: cipherText, key: decryptionKey,
                 algorithm: algorithm.ccAlgorithm,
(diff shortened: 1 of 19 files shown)


JOSESwift/Sources/CryptoImplementation/AES.swift
```diff
@@ -29,15 +29,18 @@ internal enum AESError: Error {
...
     case cannotPerformOperationOnEmptyDataBuffer
+    case invalidAlgorithm
 }
 
 fileprivate extension ContentEncryptionAlgorithm {
     var ccAlgorithm: CCAlgorithm {
         switch self {
         case .A256CBCHS512:
+            // 128 referes to the block size in this case.
             return CCAlgorithm(kCCAlgorithmAES128)
```
nathan-mohemian: As opposed to key, I suppose?

JOSESwift/Sources/CryptoImplementation/AES.swift
```diff
@@ -97,37 +123,122 @@ internal struct AES {
...
+            return plaintext
+        }
+    }
+
+    /// Encrypts the given raw key using AES key wrap.
+    /// - Parameters:
+    ///   - rawKey: The raw key to encrypt
+    ///   - keyEncryptionKey: The key used to encypt the raw key
+    ///   - algorithm: The algorithm to use for AES key wrap
+    static func keyWrap(rawKey: Data, keyEncryptionKey: Data, algorithm: KeyManagementAlgorithm) throws -> Data {
```
nathan-mohemian: Isn't this naming a little backwards? shouldn't this be `wrapKey` or `wrap(key:`? Same goes for below, I suppose.

JOSESwift/Sources/CryptoImplementation/AES.swift
```diff
@@ -97,37 +123,122 @@ internal struct AES {
...
+        switch algorithm {
+        case .A128KW, .A192KW, .A256KW:
+            guard
+                let keyLengthOk = algorithm.checkAESKeyLength(for: keyEncryptionKey),
+                keyLengthOk
```
nathan-mohemian: This can be done in one line without let binding
```swift
algorithm.checkAESKeyLength(for: keyEncryptionKey) == true
```

same applies below 👇

JOSESwift/Sources/CryptoImplementation/AES.swift
```diff
@@ -97,37 +123,122 @@ internal struct AES {
...
+            }
+
+            // See https://tools.ietf.org/html/rfc3394#section-2.2.3.1
+            // The default iv is defined to be the hexadecimal constant A6A6A6A6A6A6A6A6
+            let iv = Data(bytes: CCrfc3394_iv, count: CCrfc3394_ivLen)
+
+            let wrapped = ccAESKeyWrap(rawKey: rawKey, keyEncryptionKey: keyEncryptionKey, iv: iv)
+
+            guard
+                let wrappedKey = wrapped.data,
```
nathan-mohemian: Again, no need for binding here.
```swift 
wrapped.data?.status == kCCSuccess
```

same applies below 👇

JOSESwift/Sources/CryptoImplementation/AES.swift
```diff
@@ -97,37 +123,122 @@ internal struct AES {
...
+    }
+
+    /// Encrypts the given raw key using AES key wrap.
+    /// - Parameters:
+    ///   - rawKey: The raw key to encrypt
+    ///   - keyEncryptionKey: The key used to encypt the raw key
+    ///   - algorithm: The algorithm to use for AES key wrap
+    static func keyWrap(rawKey: Data, keyEncryptionKey: Data, algorithm: KeyManagementAlgorithm) throws -> Data {
+        switch algorithm {
+        case .A128KW, .A192KW, .A256KW:
```
nathan-mohemian: What do you think of refactoring this to a guard? It's a little clunkier, but you can drop an indentation level, and have the "else" case much closer to it's check.
```swift
let okAlgorithms = [.A128KW, .A192KW, .A256KW]
guard okAlgorithms.contains(algorithm) else {
     throw AESError.invalidAlgorithm 
}
```

JOSESwift/Sources/CryptoImplementation/RSA.swift
```diff
@@ -240,7 +241,10 @@ internal struct RSA {
 
         // Check if the plain text length does not exceed the maximum.
         // e.g. for RSA1_5 the plaintext must be 11 bytes smaller than the public key's modulus.
-        guard algorithm.isPlainTextLengthSatisfied(plaintext, for: publicKey) else {
+        guard
+            let plaintextLengthOk = algorithm.isPlainTextLengthSatisfied(plaintext, for: publicKey),
+            plaintextLengthOk
```
nathan-mohemian: Same as elsewhere, skip the binding and `== true` that sucker

also below 👇

Tests/RSADecryptionTests.swift
```diff
@@ -122,7 +124,7 @@ class RSADecryptionTests: RSACryptoTestCase {
             return
         }
 
-        for algorithm in KeyManagementAlgorithm.allCases {
+        for algorithm in keyManagementModeAlgorithms {
             guard algorithm != .direct else {
                 continue
             }
```
nathan-mohemian: Can't you omit all these `guard not direct` checks now that `keyManagementModeAlgorithms` has fixed contents?

Tests/JWERSATests.swift
```diff
@@ -92,7 +94,7 @@ class JWERSATests: RSACryptoTestCase {
             return
         }
 
-        for algorithm in KeyManagementAlgorithm.allCases {
+        for algorithm in keyManagementModeAlgorithms {
             guard algorithm != .direct else {
                 continue
             }
```
nathan-mohemian: Ditto other comment, these guards are useless now

Tests/EncrypterDecrypterInitializationTests.swift
```diff
@@ -25,131 +25,192 @@
...
+            case aesKeyManagementModeAlgorithms:
+                XCTAssertNotNil(
+                    Encrypter(
+                        keyManagementAlgorithm: algorithm,
+                        contentEncryptionAlgorithm: .A128CBCHS256,
+                        encryptionKey: Data()
+                    )
+                )
+            default:
+                // .direct
```
nathan-mohemian: Not a huge fan of twisting switch statements like this (they're really more like `if` statements in this case as the exhaustiveness guarantee isn't quite there.

Not a huge enough issue that I'd ask for a change (I get this saves a ton of boilerplate), but big enough for me to comment on.

JOSESwift/Sources/CryptoImplementation/AES.swift
daniel-mohemian: Yes exactly! i checked and `kCCAlgorithmAES128` is actually deprecated in favor of just `kCCAlgorithmAES` due to this exact ambiguity. I updated it to `kCCAlgorithmAES`.

JOSESwift/Sources/CryptoImplementation/AES.swift
daniel-mohemian: You are 💯 right. I updated it.

JOSESwift/Sources/CryptoImplementation/AES.swift
daniel-mohemian: Omg yes thank you. 😅  That's almost a bit embarrassing. 🤕

JOSESwift/Sources/CryptoImplementation/AES.swift
daniel-mohemian: In this case I reuse the let-bound `wrappedKey` later though.

nathan-mohemian: Aha, my mistake then.

JOSESwift/Sources/CryptoImplementation/AES.swift
daniel-mohemian: Good point. There might be other key wrap algorithms in the future where a switch will make sense but we don't need to care about them now. Refactored to guard. ✅

Tests/RSADecryptionTests.swift
daniel-mohemian: Good catch. Removed.

Tests/EncrypterDecrypterInitializationTests.swift
daniel-mohemian: I agree
. Updated it to be more explicit about the `.direct` case.

PR #216, opened 2020-02-20
Fine tune API changes

daniel-mohemian:
Provides deprecated API fallbacks for breaking API changes. This should allow us to release the key wrap implementation as non breaking change as part of the 2.x.x major release.


daniel-mohemian:
> Nice catch that this could be done without breaking changes!

Thanks for the typealias idea! 😄

PR #220, opened 2020-06-04
Add JWK thumbprint computation

Torsph:
Im currently using the thumbprint computed from the required fields in the key as the keyIdentifier in my own project and wanted to see if you wanted to add it to JOSESwift.

Its based on the [RFC-7638](https://tools.ietf.org/html/rfc7638).

Hopefully it can help other people also :)

Just give me a shout out if something needs to be changed or rewritten :)

### DIFF
--- a/JOSESwift/Sources/Algorithms.swift
+++ b/JOSESwift/Sources/Algorithms.swift
@@ -96,6 +96,13 @@ public enum HMACAlgorithm: String {
     }
 }
 
+/// An algorithm for Thumbprint calculation.
+///
+/// - SHA256
+public enum ThumbprintAlgorithm: String {
+    case SHA256
+}
+
 /// An algorithm for compressing the plain text before encryption.
 /// List of [supported compression algorithms](https://www.iana.org/assignments/jose/jose.xhtml#web-encryption-compression-algorithms)
 ///
--- a/JOSESwift/Sources/Algorithms.swift
+++ b/JOSESwift/Sources/Algorithms.swift
@@ -96,10 +96,10 @@ public enum HMACAlgorithm: String {
     }
 }
 
-/// An algorithm for Thumbprint calculation.
+/// An algorithm for JWK Thumbprint calculation.
 ///
 /// - SHA256
-public enum ThumbprintAlgorithm: String {
+public enum JWKThumbprintAlgorithm: String {
     case SHA256
 }
 
--- a/JOSESwift/Sources/CryptoImplementation/Thumbprint.swift
+++ b/JOSESwift/Sources/CryptoImplementation/Thumbprint.swift
@@ -1,0 +1,68 @@
+//
+//  SHA256.swift
+//  JOSESwift
+//
+//  Created by Thomas Torp on 08.06.20.
+//
+//  ---------------------------------------------------------------------------
+//  Copyright 2019 Airside Mobile Inc.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//  ---------------------------------------------------------------------------
+//
+
+import Foundation
+import CommonCrypto
+
+enum ThumbprintError: Error {
+    case inputMustBeGreaterThanZero
+}
+
+fileprivate extension ThumbprintAlgorithm {
+    var outputLenght: Int {
+        switch self {
+        case .SHA256:
+            return Int(CC_SHA256_DIGEST_LENGTH)
+        }
+    }
+
+    func calculate(input: UnsafeRawBufferPointer, output: UnsafeMutablePointer<UInt8>) {
+        switch self {
+        case .SHA256:
+            CC_SHA256(input.baseAddress, CC_LONG(input.count), output)
+        }
+    }
+}
+
+internal struct Thumbprint {
+    /// Calculates a hash of an input with a specific hash algorithm.
+    ///
+    /// - Parameters:
+    ///   - input: The input to calculate a hash for.
+    ///   - algorithm: The algorithm used to calculate the HMAC.
+    /// - Returns: The calculated hash in base64URLEncoding.
+    static func calculate(from input: Data, algorithm: ThumbprintAlgorithm) throws -> String {
+        guard input.count > 0 else {
+            throw ThumbprintError.inputMustBeGreaterThanZero
+        }
+
+        let hashBytes = UnsafeMutablePointer<UInt8>.allocate(capacity: algorithm.outputLenght)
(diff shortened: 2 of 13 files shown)


Torsph:
Forgot to fix the JWSHeader implementation. Will recreate the PR when that is done :)

daniel-mohemian:
Awesome thanks for the contribution @Torsph! It's great to see that you also added some tests for your implementation. 😄 

I'll have a more detailed look at your changes next week. Hope that's fine with you.

Torsph:
@daniel-mohemian Thanks! Sure take your time :)
My concerns is:
- Should the sha256 function be put somewhere else?
- The best way to add the thumbprint as the kid. Was wondering if I should put it in a init, but that could potentially break code written by users of the SDK. So thats why I ended up with the "withKeyIdFromThumbprint" function

JOSESwift/Sources/Algorithms.swift
```diff
@@ -96,6 +96,13 @@ public enum HMACAlgorithm: String {
     }
 }
 
+/// An algorithm for Thumbprint calculation.
```
daniel-mohemian, line 99:
```suggestion
/// An algorithm for JWK thumbprint calculation.
```

JOSESwift/Sources/Algorithms.swift
```diff
@@ -96,6 +96,13 @@ public enum HMACAlgorithm: String {
...
+///
+/// - SHA256
+public enum ThumbprintAlgorithm: String {
```
daniel-mohemian, line 102:
```suggestion
public enum JWKThumbprintAlgorithm: String {
```

To be able to more clearly distinguish this from `HMACAlgorithm` it might make sense to add the `JWK` prefix.

JOSESwift/Sources/CryptoImplementation/Thumbprint.swift
```diff
@@ -0,0 +1,68 @@
...
+        }
+    }
+}
+
+internal struct Thumbprint {
+    /// Calculates a hash of an input with a specific hash algorithm.
+    ///
+    /// - Parameters:
+    ///   - input: The input to calculate a hash for.
+    ///   - algorithm: The algorithm used to calculate the HMAC.
```
daniel-mohemian, line 52:
```suggestion
    ///   - algorithm: The algorithm used to calculate the hash.
```

JOSESwift/Sources/CryptoImplementation/Thumbprint.swift
```diff
@@ -0,0 +1,68 @@
...
+        switch self {
+        case .SHA256:
+            return Int(CC_SHA256_DIGEST_LENGTH)
+        }
+    }
+
+    func calculate(input: UnsafeRawBufferPointer, output: UnsafeMutablePointer<UInt8>) {
+        switch self {
+        case .SHA256:
+            CC_SHA256(input.baseAddress, CC_LONG(input.count), output)
```
daniel-mohemian, line 42: Would it make sense to check the return type here to make sure the hash was calculated properly?

JOSESwift/Sources/CryptoImplementation/Thumbprint.swift
```diff
@@ -0,0 +1,68 @@
...
+    /// - Returns: The calculated hash in base64URLEncoding.
+    static func calculate(from input: Data, algorithm: ThumbprintAlgorithm) throws -> String {
+        guard input.count > 0 else {
+            throw ThumbprintError.inputMustBeGreaterThanZero
+        }
+
+        let hashBytes = UnsafeMutablePointer<UInt8>.allocate(capacity: algorithm.outputLenght)
+        defer { hashBytes.deallocate() }
+
+        input.withUnsafeBytes { (buffer) -> Void in
```
daniel-mohemian, line 62:
```suggestion
        input.withUnsafeBytes { buffer in
```

JOSESwift/Sources/DataExtensions.swift
```diff
@@ -22,6 +22,7 @@
 //
 
 import Foundation
+import CommonCrypto
```
daniel-mohemian, line 25:
```suggestion
```

JOSESwift/Sources/JWK.swift
```diff
@@ -83,4 +85,25 @@ public protocol JWK: Codable {
     /// - Returns: The JSON representation of the JWK as `Data` or
     ///            `nil` if the encoding failed.
     func jsonData() -> Data?
+
+    /// Generate the thumbprint based on hash function SHA256
+    ///  See [RFC-7638, Section 3.2](https://tools.ietf.org/html/rfc7638#section-3.2)
+    ///
+    /// - Returns: The thumbprint from the required members of the JWK key.
+    /// - Throws: A `JOSESwiftError` indicating any errors.
```
daniel-mohemian, line 93:
```suggestion
    /// Generate the key's thumbprint.
    ///  See [RFC-7638](https://tools.ietf.org/html/rfc7638).
    ///
    /// - Parameters:
    ///   - algorithm: The hash algorithm to use for the thumbprint calculation.
    /// - Returns: The base64url encoded thumbprint of the required members of the JWK key.
    /// - Throws: A `JOSESwiftError` indicating any errors.
```

JOSESwift/Sources/ECKeys.swift
```diff
@@ -187,6 +197,14 @@ public struct ECPublicKey: JWK {
 
         return try T.representing(ecPublicKeyComponents: (self.crv.rawValue, x, y))
     }
+
+    @available(iOS 11.0, *)
+    public func withKeyIdFromThumbprint(algorithm: ThumbprintAlgorithm = .SHA256) throws -> Self {
```
daniel-mohemian, line 202: We need to add a documentation comment to the `withKeyIdFromThumbprint` functions. To avoid having to add the comment to all versions of the function you could make this an internal helper function and then call the helper functions once in a public extension to JWK.

JOSESwift/Sources/JWK.swift
```diff
@@ -83,4 +85,25 @@ public protocol JWK: Codable {
...
+    @available(iOS 11.0, *)
+    func thumbprint(algorithm: ThumbprintAlgorithm) throws -> String
+
+    @available(iOS 11.0, *)
+    func withKeyIdFromThumbprint(algorithm: ThumbprintAlgorithm) throws -> Self
```
daniel-mohemian, line 98: Maybe something like this would be clearer: 

```suggestion
    func withThumbprintAsKeyId(algorithm: ThumbprintAlgorithm) throws -> Self
```

What do you think?

Torsph: Yes, totally agree :)

JOSESwift/Sources/ECKeys.swift
Torsph: Pushed some changes now :)

JOSESwift/Sources/CryptoImplementation/Thumbprint.swift
Torsph: I have never actually validated the return value from CC_SHA256.
Based on the man page it just return the pointer passed in as the md parameter. 
```
extern unsigned char *
     CC_SHA256(const void *data, CC_LONG len, unsigned char *md);
```

Torsph:
Perfect! Thanks for the help!

daniel-mohemian:
@Torsph your changes are now live as part of https://github.com/airsidemobile/JOSESwift/releases/tag/2.2.0.

PR #225, opened 2020-06-23
Copy parameters when updating JWT with keyId

Torsph:
Hi again!

I found a minor bug when using the withThumbprintAsKeyId function. I forgot to copy potensial additionalParameters 😬 
Added the fix and a couple of new tests.


daniel-mohemian:
Whoops my mistake, I should have seen that. Thanks for fixing! It's released as 2.2.1.

Torsph:
Perfect! Thanks :)

PR #231, opened 2020-08-13
Support for RSAPrivateKey init from JWK token

lukereichold:
Sharing my implementation for initializing an `RSAPrivateKey` from a JWK token.

NOTE: with these additions, the legacy tests around `RSAPrivateKey` are failing, as this implementation requires all RSA fields to be provided (i.e. prime1, prime2, exponent1, exponent2, and coefficient) in addition to just `privateExponent`. I currently don't have the bandwidth to address these, but moving forward to remedy this I recommend adding support for initializing an `RSAPrivateKey` with only `modulus`, `exponent`, and `privateExponent`.

#helpWanted

### DIFF
--- a/.swiftpm/xcode/package.xcworkspace/contents.xcworkspacedata
+++ b/.swiftpm/xcode/package.xcworkspace/contents.xcworkspacedata
@@ -1,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<Workspace
+   version = "1.0">
+   <FileRef
+      location = "self:">
+   </FileRef>
+</Workspace>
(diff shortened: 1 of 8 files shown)


Jonathan-Airside:
Hey @lukereichold, thanks for opening a PR!

Were all pretty busy here too, but when someone has a minute well review your feature and help resolve any tests!

daniel-mohemian:
Hey @lukereichold thanks for the contribution! It would definitely be cool to support these additional parameters for private keys so thanks for kicking that off.

As you mentioned, we definitely need support for initializing an `RSAPrivateKey` with only modulus, exponent, and private exponent. This is also reflected in [RFC-7518, 6.3](https://tools.ietf.org/html/rfc7518#section-6.3.2):

> In addition to the members used to represent RSA public keys, the following members are used to represent RSA private keys.  The parameter "d" is REQUIRED for RSA private keys.  The others enable optimizations and SHOULD be included by producers of JWKs representing RSA private keys.  If the producer includes any of the other private key parameters, then all of the others MUST be present, with the exception of "oth", which MUST only be present when more than two prime factors were used.

How about making the additional parameters (p, q, dp, dq, qi) optional? Of course we'd need to ensure that if one additional parameter is present, all others are present as well. That should also solve the issues with the old tests - obviously requiring a set of new test covering the case of private keys with such additional parameters.

One thing I'm not sure about is if `SecKeyCreateWithData` will accept private key data containing only n, e, and d though. If it only accepts key data including the additional parameters - that would be a problem. In that case we might have to think about dropping support for keys which only contain n, e, and d. That would be non-ideal though. 😅 Do you know more about the behavior of `SecKeyCreateWithData` in that case?

daniel-mohemian:
I'm going to close this PR for now because it hasn't been updated in quite a while. Anyone feel free to reopen as required.

PR #240, opened 2020-10-30
Add parsed JWK header parameter

johanfylling:
The JWS and JWE specifications dictate that the 'jwk' header parameter should be represented as a JWK (https://tools.ietf.org/html/rfc7515#section-4.1.3 & https://tools.ietf.org/html/rfc7516#section-4.1.5). The JWK representation is a JSON dictionary, and not a string, as currently implemented by JOSESwift.

This pull request is a suggested solution. It is also a work in progress, as the implementation of the 'jwk' header getters is sub-optimal and a better approach may be more desirable.

This pull request also does not provide test coverage, and short-circuits some tests in benefit of successful compilation.


daniel-mohemian:
Hi @johanfylling 👋  Good find! That indeed looks like a mismatch between implementation and spec. We'll have a look at you pull request soon and will get back to you with some feedback.

daniel-mohemian:
Hi @johanfylling! 

Sorry for the delay. Your implementation looks pretty solid. Our next steps on this pull request would be to get the tests fully up and running again. Also I think we could speed up the release of this if we'd keep around the old `String` parameter for compatibility reasons. I'm thinking let the old `jwk` parameter be a `String` and add a new computed one that transforms the `String` to a `JWK` (just as you implemented it already).

Since that way we wouldn't introduce any breaking API changes, we could release this much faster as a minor version bump and wouldn't have to wait for the next major release. Moreover, it would make it much easier for you to fix the tests as the old tests should just run ok again and you would only have to add one or two test cases that cover the new parsed JWK parameter. 

Let me know what you think!

johanfylling:
I have updated the pull request to now have two parameters that both read/write to the 'jwk' header param.
The 'jwk' parameter now is a String optional, and behaves just as before. The new 'jwkTyped' parameter assumes the 'jwk' header parameter to be a dictionary, and if used as a setter, a dictionary 'jwk' will be produced for the serialization.
Is this an acceptable solution? If so, I can spend some time to write tests to ensure this parameter behaves as advertised.

JOSESwift/Sources/JWEHeader.swift
```diff
@@ -154,6 +154,36 @@ extension JWEHeader: CommonHeaderParameterSpace {
...
+        }
+        get {
+            guard let jwkParameters = parameters["jwk"] as? [String: String] else {
+                return nil
+            }
+            
+            guard let keyTypeString = jwkParameters[JWKParameter.keyType.rawValue],
+                  let keyType = JWKKeyType(rawValue: keyTypeString) else {
+                return nil
+            }
```
daniel-mohemian, line 170:
```suggestion
            guard 
                let keyTypeString = jwkParameters[JWKParameter.keyType.rawValue],
                let keyType = JWKKeyType(rawValue: keyTypeString)
            else {
                return nil
            }
```

JOSESwift/Sources/JWSHeader.swift
```diff
@@ -103,8 +103,8 @@ extension JWSHeader: CommonHeaderParameterSpace {
             return URL(string: parameter)
         }
     }
-
-    /// The JSON Web key corresponding to the key used to digitally sign the JWS.
+
```
daniel-mohemian, line 106:
```suggestion

```

JOSESwift/Sources/JWSHeader.swift
```diff
@@ -114,6 +114,36 @@ extension JWSHeader: CommonHeaderParameterSpace {
...
+        }
+        get {
+            guard let jwkParameters = parameters["jwk"] as? [String: String] else {
+                return nil
+            }
+            
+            guard let keyTypeString = jwkParameters[JWKParameter.keyType.rawValue],
+                  let keyType = JWKKeyType(rawValue: keyTypeString) else {
+                return nil
+            }
```
daniel-mohemian, line 130:
```suggestion
            guard
                let keyTypeString = jwkParameters[JWKParameter.keyType.rawValue],
                let keyType = JWKKeyType(rawValue: keyTypeString)
            else {
                return nil
            }
```

daniel-mohemian (commented):
Yes this would definitely be an acceptable solution. The duplication is a bit unfortunate but this is our fault and we'll clean that up once we get a chance to redo the header implementation.

If you have time to add some more test that hit both the JWE and JWS implementation - that would be greatly appreciated.

johanfylling:
I've added tests for JWE- and JWS headers. The two failing tests were already failing on the tag (2.2.1) I branched off from.

daniel-mohemian:
Marking this PR ready for review now.

daniel-mohemian (approved):

daniel-mohemian:
I'll let you know once this is released @johanfylling. Thanks for the contribution!

johanfylling:
Thank you very much for your input @daniel-mohemian and for accepting our proposal!

daniel-mohemian:
Hey @johanfylling 👋 Your contribution is now live as [2.3.0](https://github.com/airsidemobile/JOSESwift/releases/tag/2.3.0).

johanfylling:
Awesome! Thank you very much for a speedy release.

PR #243, opened 2020-11-16
Build maintenance

daniel-mohemian:
- Bump executor to fix an issue with Sonarqube (it now seems to require Java 11 which wasn't available on the previous executor).
- Automatically create the release preparation pull request when running the prepare lane.
- Fix linter issues.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@ version: 2.1
 executors:
   mac-executor:
     macos:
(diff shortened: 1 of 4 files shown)


daniel-mohemian:
The duplication issue is known and is because of the basically equivalent implementations of `JWEHeader` and `JWSHeader`. We should fix this sometime but it is unrelated to this pull request.

haeser (approved):

PR #246, opened 2020-12-13
Stop installing SwiftLint when it's not installed

barakwei:
The SwiftLint script run phase used backticks around "brew bundle" when
printing a message when SwiftLint is not installed. In shell,
expressions in backticks are evaluated, or in other words, executed.
So every time this script ran on a machine without Swiftlint installed
it basically install SwiftLint. This is not a desired behavior.


daniel-mohemian (approved):
You're totally right. Thanks for the fix @barakwei! ✅

daniel-mohemian:
Released as [2.3.1](https://github.com/airsidemobile/JOSESwift/releases/tag/2.3.1).

PR #249, opened 2020-12-14
Automate publishing releases on GitHub

daniel-mohemian:
This will automatically publish a release including the latest changelog on GitHub when the `release` lane is run.

The latest release ([2.3.1](https://github.com/airsidemobile/JOSESwift/releases/tag/2.3.1)) was created using this lane.

Running the whole `release` lane on CI whenever a prepare pr is merged would be the next step in automating the release process.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -87,5 +87,3 @@ fastlane/test_output
 # Sonarqube
 .scannerwork
 
-# Vi
-.swp
--- a/.gitignore
+++ b/.gitignore
@@ -86,3 +86,6 @@ fastlane/test_output
 
 # Sonarqube
 .scannerwork
+
+# Vi
+.swp
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -174,7 +174,7 @@ lane :publish_release do
   version = get_version_number(target: "JOSESwift")
 
   changelog = File.read("../CHANGELOG.md")
-  changes = changelog[/(- .+\(#.+\)\n)+/]
+  changes = changelog[/(- .+\(#.+\)\n)+/] # Filter latest change set (e.g. "- Add some feature (#123)\n...").
   tmp_changelog = Tempfile.new("jose-changelog")
   tmp_changelog.write(changes)
   tmp_changelog.close
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -1,4 +1,5 @@
-require 'fileutils'
+require "fileutils"
+require "tempfile"
 
 fastlane_version "2.63.0"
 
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -86,3 +86,6 @@ fastlane/test_output
 
 # Sonarqube
 .scannerwork
+
+# Vi
+.swp
```
nathan-mohemian, line 91: Isn't this file generated anyway? I bet they support vim too.

fastlane/Fastfile
```diff
@@ -162,9 +163,27 @@ lane :release do
...
+
   ensure_git_status_clean
 end
 
+desc "Publishes the release on GitHub"
+lane :publish_release do
+  version = get_version_number(target: "JOSESwift")
+
+  changelog = File.read("../CHANGELOG.md")
+  changes = changelog[/(- .+\(#.+\)\n)+/]
```
nathan-mohemian, line 177: What is this regex doing? I get you're stripping chars, but why exactly?

.gitignore
nathan-mohemian: Alternatively, does it make sense to commit this at all? Would we accept a PR for emacs support in our gitignore?

Maybe this is something users should configure locally, like so? https://stackoverflow.com/a/5724476

daniel-mohemian: Since its mostly me working on this code I quickly setup the gitignore to fit my working environment. User level global gitignores would be a good way of solving this as well. Probably superior even. 👍  

If someone regularly commits to the project and wants an emacs gitignore I'd be happy to accept a PR. 😅

fastlane/Fastfile
daniel-mohemian: Filtering out the latest change set.

``` 
Blablabla

[1.2.0]

- Do Something (#123)
- Do Something else (#100)

[1.2.0]

- Start something (#99)
```

Will become:

```
- Do Something (#123)
- Do Something else (#100)
```

Which is the change log for the latest version that we want attached to the release.

.gitignore
nathan-mohemian: Fair enough

fastlane/Fastfile
nathan-mohemian: Might be worth dropping a source comment

nathan-mohemian (approved):

fastlane/Fastfile
daniel-mohemian: Yeah you're right. 👍  Will add this later!

daniel-mohemian:
@nathan-mohemian Updated .gitignore and Fastfile. 👌

nathan-mohemian (approved):

Issue #266, opened 2022-01-03
Support for library evolution


byohay:
Hi,
We've been using `JOSESwift` for some time, and we want to start precompiling to save build time. I know that this repository doesn't provide binaries, but it would still be helpful for us if this project would be guaranteed to compile with `BUILD_LIBRARY_FOR_DISTRIBUTION` enabled. Otherwise we might have to diverge from your fork to make it compile.
In practice this probably means creating tests that compile with this flag and check that the build succeeds.
The same was done in [GRDB](https://github.com/groue/GRDB.swift/pull/1037).

byohay:
@nathan-mohemian thanks! I opened a PR #267

nathan-mohemian:
Glad to here it! Will take a look tomorrow.

PR #267, opened 2022-01-10
Fastfile: Run tests with `BUILD_LIBRARY_FOR_DISTRIBUTION`.

byohay:
Solves #266

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -14,7 +14,7 @@ lane :test do
   sh "rm -rf ../#{DERIVED_DATA_PATH}"
 
   UI.message "Testing"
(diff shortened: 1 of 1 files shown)


nathan-mohemian:
Is the intent here just to make sure that running tests with `BUILD_LIBRARY_FOR_DISTRIBUTION` doesn't break anything? I would have expected some kind of change in the project file or package spec as well.

byohay:
From my understanding, if your clients don't intend to build JOSESwift separately you shouldn't enable library evolution (See [here](https://www.swift.org/blog/library-evolution/)). So if your clients wants to build it separately they should enable it themselves.
On another matter: Clients who pre-compile the library will use the `archive` command, but the current check runs the tests. Should I add another step that checks with the `archive` command? Or is it an overkill?

PR #270, opened 2022-04-27
Add support for JWS unencoded detached payloads

visav-tietoevry:
## Summary

Add option to skip base64url encoding of payload in JWS. This results in a compact serialization with [detached payload](https://www.rfc-editor.org/rfc/rfc7515.html#appendix-F).

Motivation for not encoding the payload is further described in [RFC-7797](https://datatracker.ietf.org/doc/html/rfc7797#section-1) _JSON Web Signature (JWS) Unencoded Payload Option_.

One example use case where unencoded payload is needed is implementing [`JsonWebSignature2020`](https://w3c-ccg.github.io/lds-jws2020/#suite-definition) for JSON-LD document proofs. In this case, JWS is used as a signature, and the document itself is not embedded in the JWS. This suite takes JSON-LD document & it's proof without signature, does some steps to normalize them, computes SHA-256 digests, concatenates them and forms a JWS from this result, without using base64url encoding. Finally, resulting JWS is added to the document proof segment, without payload. Verifying end can then reproduce the steps to create the payload for JWS and verify the signature.

## Changes

- Add [`"b64"`](https://datatracker.ietf.org/doc/html/rfc7797#section-3) header parameter support to `JWSHeader` to enable unencoded payload
- Replace `asJOSESigningInput()` with `JWSSigningInput` struct, which now contains the logic for signing input, controlled by `"b64"` header parameter
- Add support for initializing `JWS` with detached payload, to enable verification of (unencoded) detached payloads

## Caveats

- Due to [Unencoded Payload Content Restrictions](https://datatracker.ietf.org/doc/html/rfc7797#section-5), compact serialization of `JWS` with `"b64"` header parameter always yields a serialization with a detached payload. This was a choice to limit the scope of the PR to get minimum viable support for unencoded payload option.

## Remarks

Following tests were failing for me at `master` (d542b31):

- `testDecryptingAliceSecretWithBobKey()`
- `testDecryptingBobSecretWithAliceKey()`

These test failures are not related to changes in this PR.

## Tests

I found it a bit difficult to find a proper home for the unit tests of the added JWS functionality, so I chose to make a new test case class, `JWSUnencodedPayloadTests`, which was arbitrarily inherited from `ECCryptoTestCase` to get the required dependencies for signing & verifying.

I think these unit tests should be agnostic of the signature algorithm, but I couldn't find an easy way to do so without doing some bigger refactoring. If there's a simple way to mock `Signer`, please let me know.

### DIFF
--- a/JOSESwift.xcodeproj/project.pbxproj
+++ b/JOSESwift.xcodeproj/project.pbxproj
@@ -7,6 +7,7 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
+		5C0F564327FF18C8006328D1 /* JWSSigningInput.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5C0F564227FF18C8006328D1 /* JWSSigningInput.swift */; };
 		5FB760497BB7711EFB470B5A /* ECKeys.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5FB76A41AECF99F3673C1C40 /* ECKeys.swift */; };
 		5FB76093CE81BF1F8E7C254A /* JWKECDecodingTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5FB7625CF5EF5D8F2135967F /* JWKECDecodingTests.swift */; };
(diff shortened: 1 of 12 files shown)


daniel-moh:
Hi @visav-tietoevry thanks for opening this PR and sorry for the late reply. We've just re-booted the project and released a new 3.0.0 release. I will have a look at your PR in the coming weeks. We're planning to include it in the next release.

Issue #251, opened 2021-01-11
Architecture for external dependencies


mitchins:
Apologies for the lengthiness, there's a fair bit to compress:

I am working for a digital bank and we use this library, the only caveat is that AES256GCM isn't supported.
We've had our upstream vendor give us a hand and it looks like we would be able to expand this library to supported it...

Now ideally I'd like to contribute this back upstream and not keep it in a silo.
Here comes the gotcha: CryptoKit is only supported on iOS >= 13.0, and many companies/people still support 12 or even 11 due to the number of devices still out there such as iPhone 6 etc.

Long story short, this functionality - even if optional would require https://github.com/krzyzanowskim/CryptoSwift.
Now the tricky bit:
* Currently there are no external dependencies for the library building itself.
* JOSESwift is supporting both Carthage and CocoaPods. (So does CryptoSwift)

Here's my thinking....
We don't want to force anything upon a user but if they want AES256GCM we'd need to add this library, the library would be needed.

I know for a fact cocoa pods wouldn't require the project files as it constructs a new one based on the pod spec.
I am less familiar with Carthage but believe that it will use the existing xcproject file, and so the current project.

Here are the two **viable** options as far as I can see it:
1. As a direct dependency:
CryptoSwift module is added as a dependency for this JOSESwift via pod and Carthage respectively. Now if someone was to open up the project itself and build it from Xcode it would need a (1) pod or (2) cartfile to install the module, or swift package manager. My only holdout here is that the project needs to utilise one of the 3 package managers if someone directly opens the project (pod of course won't care since it creates a new one from scratch every time based on the spec).

2. As an injectable dependency
This would be more work for the client perhaps, but would alleviate changing dependencies.
I imagine this JoseSwift exposes a protocol for what it needs to be like `GenericEncrypterStrategy` 

In order to actually use AES256GCM, we'd need to fulfil the dependency:
There is both `JWEHeader` and `Encrypter`, which we'd either:
a) Inject the encryption provider per instance (in the constructor most likely)
b) Inject it into the module via some static property (`JoseSwift.provider = Foo.Bar` or something)

While one needs to be careful with static stuff, at a module/library level is possibly  _ok_ (option b). The first option (a) would lead to longer initialisation etc. 

It would also be possible to implement a default provider of CryptoKit allowing iOS >= 13.0 to not do any work.
Do not have this as of yet, but it would merely conform to the expected protocol for a `GenericEncrypterStrategy `.

[...]

PR #278, opened 2022-08-12
[JWE] Added support for `A256GCM` and `A128GCM`

tobihagemann:
After considering the comments in #251, I've added support for content encryption using AES GCM according to JWE standard. As I've mentioned there, I increased the minimum iOS target to 13.0 because of CryptoKit.


istarp:
Thanks for this PR, any thoughts on completing it and releasing new lib version?

tobihagemann:
From my perspective, it's completed. I'm just waiting for a review.

istarp:
I hope someone will take a look, seems like most of the PR's are not reviewed at all :(

kiransable279:
> After considering the comments in #251, I've added support for content encryption using AES GCM according to JWE standard. As I've mentioned there, I increased the minimum iOS target to 13.0 because of CryptoKit.

Any update on this. Please include AES128GCM content encryption algorithm. It is very important and urgent. I dont understand why this is not included in the library?

lsolniczek:
Hi, any update on this? It will be very useful to have this PR merged.

Thireus:
Looks like it fails lint tests - @tobihagemann

sfostermc:
Is there anything we can do to help get this PR merged in?

JOSESwift/Sources/AESCBCEncryption.swift
```diff
@@ -26,29 +26,48 @@ import Foundation
...
+        }
+        // now select the appropriate hashing algorithm
+        switch contentEncryptionAlgorithm {
+        case .A256CBCHS512:
+            hmacAlgorithm = .SHA512
+        case .A128CBCHS256:
+            hmacAlgorithm = .SHA256
+        case .A256GCM, .A128GCM:
+            throw JWEError.contentEncryptionAlgorithmMismatch
+        }
```
daniel-moh, line 56: Totally a taste thing, not a change request:

Should we move these two switches to separate functions just like the `getAuthenticationTag` below? For me personally I feel like it's easier to keep an overview that way. But Both options work for me. Again, it's a taste thing.

JOSESwift/Sources/AESGCMEncryption.swift
```diff
@@ -0,0 +1,67 @@
...
+        self.contentEncryptionKey = contentEncryptionKey
+    }
+
+    func encrypt(_ plaintext: Data, additionalAuthenticatedData: Data) throws -> ContentEncryptionContext {
+        let iv = try SecureRandom.generate(count: contentEncryptionAlgorithm.initializationVectorLength)
+        return try encrypt(plaintext, initializationVector: iv, additionalAuthenticatedData: additionalAuthenticatedData)
+    }
+
+    func encrypt(_ plaintext: Data, initializationVector: Data, additionalAuthenticatedData: Data) throws -> ContentEncryptionContext {
+        return try AESGCM.encrypt(plaintext: plaintext, encryptionKey: contentEncryptionKey, initializationVector: initializationVector, additionalAuthenticatedData: additionalAuthenticatedData)
```
daniel-moh, line 41: Should we split these calls to multiple lines for easier readability?

```suggestion
        return try AESGCM.encrypt(
            plaintext: plaintext,
            encryptionKey: contentEncryptionKey,
            initializationVector: initializationVector,
            additionalAuthenticatedData: additionalAuthenticatedData
        )
```

JOSESwift/Sources/CryptoImplementation/AESGCM.swift
```diff
@@ -0,0 +1,60 @@
...
+        initializationVector: Data,
+        additionalAuthenticatedData: Data
+    ) throws -> ContentEncryptionContext {
+        let key = CryptoKit.SymmetricKey(data: encryptionKey)
+        let nonce = try CryptoKit.AES.GCM.Nonce(data: initializationVector)
+        let encrypted = try CryptoKit.AES.GCM.seal(plaintext, using: key, nonce: nonce, authenticating: additionalAuthenticatedData)
+        return ContentEncryptionContext(
+            ciphertext: encrypted.ciphertext,
+            authenticationTag: encrypted.tag,
+            initializationVector: encrypted.nonce.withUnsafeBytes({ Data(Array($0)) })
```
daniel-moh, line 33: Is 

``` swift
encrypted.nonce.withUnsafeBytes({ Data(Array($0)) })
```

the same as

``` swift
initializationVector
```

?

daniel-moh (commented):
Very nice! Couple of comments, let me know what you think!

JOSESwift/Sources/AESCBCEncryption.swift
```diff
@@ -26,29 +26,48 @@ import Foundation
...
+        guard contentEncryptionAlgorithm.checkKeyLength(for: contentEncryptionKey) else {
+            throw JWEError.keyLengthNotSatisfied
+        }
         self.contentEncryptionAlgorithm = contentEncryptionAlgorithm
-        self.contentEncryptionKey = contentEncryptionKey
+        // the key comes in as a combined key, split it according to the algorithm
+        switch contentEncryptionAlgorithm {
+        case .A256CBCHS512:
+            self.hmacKey = contentEncryptionKey.subdata(in: 0..<32)
+            self.contentEncryptionKey = contentEncryptionKey.subdata(in: 32..<64)
```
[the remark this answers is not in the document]
haeser: Very valid and valuable feedback. I like the naming secret key for the combined one and will use macKey and encryptionKey for the split ones. Will make the adaptions.

JOSESwift/Sources/CryptoImplementation/AESGCM.swift
haeser: Yes it is. This way all the context was built out of encrypted output. 
Changed it to `initializationVector` for readability now. Also involves fewer conversions.

JOSESwift/Sources/AESCBCEncryption.swift
haeser: externalised both of them to functions. I like it better as well.

JOSESwift/Sources/AESGCMEncryption.swift
haeser: done for both calls in the class

JOSESwift/Sources/CryptoImplementation/AESGCM.swift
daniel-moh: > This way all the context was built out of encrypted output.

That somehow sounds better to me I must admit. Just to double check, does it have any relevance at all if we use the incoming iv or the one from the encryption output?

JOSESwift/Sources/AESCBCEncryption.swift
```diff
@@ -63,46 +69,69 @@ struct AESCBCEncryption {
...
+                return .SHA512
+            case .A128CBCHS256:
+                return .SHA256
+            case .A256GCM, .A128GCM:
+                throw JWEError.contentEncryptionAlgorithmMismatch
+            }
+    }
+
+    static func retrieveKeys(for contentEncryptionAlgorithm: ContentEncryptionAlgorithm,
+                             from inputKey: Data) throws -> (hmacKey: Data, encryptionKey: Data) {
```
daniel-moh, line 114:
```suggestion
    static func retrieveKeys(
        for contentEncryptionAlgorithm: ContentEncryptionAlgorithm,
        from inputKey: Data
    ) throws -> (hmacKey: Data, encryptionKey: Data) {
```

daniel-moh: IIRC this is the formatting we were using in other parts of the codebase

daniel-moh (approved):
🎉

JOSESwift/Sources/AESCBCEncryption.swift
haeser: 😅 I wonder if I can install a formatter to match styles automatically.
Is there something you would recommend? 
We could go with https://github.com/apple/swift-format

daniel-moh: Yeah we should definitely automate this. IIRC we did some research in the past but I'm sure things changed in the meantime. Defaulting to Apple's formatter surely isn't a bad idea. Probably what Xcode uses under the hood too (I guess) so that might make it easier as well.

Issue #292, opened 2023-05-22
carthage build failing due to Swiftlint warning


antonyalkmim:
I'm trying to build JOSeSwift through carthage and I got this error:

```
note: Run script build phase 'SwiftLint' will be run during every build because the option to run the script phase "Based on dependency analysis" is unchecked. (in target 'JOSESwift' from project 'JOSESwift')
** ARCHIVE FAILED **


The following build commands failed:
	PhaseScriptExecution SwiftLint /Users/antony/Library/Caches/org.carthage.CarthageKit/DerivedData/14.3_14E222b/JOSESwift/2.4.0/Build/Intermediates.noindex/ArchiveIntermediates/JOSESwift/IntermediateBuildFilesPath/JOSESwift.build/Release-iphoneos/JOSESwift.build/Script-4E06574B223CC78D00D7A199.sh (in target 'JOSESwift' from project 'JOSESwift')
```

The errors its happening because this tuple has too many params: https://github.com/airsidemobile/JOSESwift/blob/f308de16698a18d74d93779bca4d702ba9186d7a/JOSESwift/Sources/ECKeys.swift#L46-L51

## Possible Solutions

- Fix SwitLint errors by disabling `large_tuple` rule or increase params limit. I've tried to remove this rule and it worked!
- Refact that code fixing the warning.
- I tried to use [this approach](https://github.com/flutter/flutter/issues/108321) adding `alwaysOutOfDate = 1;` to `.pbxproj` but it doesnt work (i dont know why 😅, maybe I should try again) 

Let me know which solution is the best here then we can fix it asap. (our C.I. its not building due to this error)

## Similar issues

https://github.com/ReactiveX/RxSwift/pull/2490
https://github.com/flutter/flutter/issues/108321

PR #293, opened 2023-05-22
Fix linter warnings

antonyalkmim:
fixes: #292 

## Description

Fix carthage build error by setting Swiftlint `large_tuple` rule to pop an error only when we have more than 5 params in a tuple.

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -18,6 +18,9 @@ disabled_rules:
   - todo
   - no_space_in_method_call
 
+large_tuple: 
+  error: 5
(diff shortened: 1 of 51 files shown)


antonyalkmim:
I tried to disable completely the `large_tupe` rule but the CI keeps failing. 

Notice that its failing because there are too many lint warnings and CI runs on strict mode.

daniel-moh (commented):
✅ LGTM, thanks for the fixes.

daniel-moh (approved):

PR #295, opened 2023-06-29
Support dynamic SPM product

andrii-vysotskyi-cko:
This PR adds a dynamic product to `JOSESwift`.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -5,7 +5,8 @@ let package = Package(
     name: "JOSESwift",
     platforms: [.iOS(.v13), .macOS(.v10_15), .watchOS(.v8), .tvOS(.v15), .visionOS(.v1)],
     products: [
-        .library(name: "JOSESwift", targets: ["JOSESwift"])
+        .library(name: "JOSESwift", targets: ["JOSESwift"]),
+        .library(name: "JOSESwift-Dynamic", type: .dynamic, targets: ["JOSESwift"])
     ],
     dependencies: [],
     targets: [


daniel-moh:
Hi @andrii-vysotskyi-cko, thanks for your contribution and apologies for the [late response](https://github.com/airsidemobile/JOSESwift/issues/305).

Could you clarify a bit on the need for declaring this separate library product? According to SPM documentation, leaving the linking type empty lets SPM decide which one to use—effectively supporting both linkage types. I might be missing something here but, to me, this sounds like there is no need to explicitly declare a separate dynamic-only library product:

> Leave this parameter so Swift Package Manager can choose between static or dynamic linking (recommended). If you don’t support both linkage types, use [Product.Library.LibraryType.static](https://developer.apple.com/documentation/packagedescription/product/library/librarytype/static) or [Product.Library.LibraryType.dynamic](https://developer.apple.com/documentation/packagedescription/product/library/librarytype/dynamic) for this parameter.

📚 [SPM Docs](https://developer.apple.com/documentation/packagedescription/product/library(name:type:targets:))

andrii-vysotskyi-cko:
Hey @daniel-moh, thank you for the response. 

The use case we are facing is that there is a binary only dynamic framework that is distributed via SPM. This framework depends on `JOSESwift`, but SPM thinks that linking of JOSESwift should be static.

Having an ability to use "dynamic"  version of `JOSESwift` explicitly should solve the problem.

daniel-moh:
Thanks for the clarification @andrii-vysotskyi-cko. I understand that adding an additional explicit dynamic library product represents the workaround you need, but it does introduce some complexity that feels a bit unconventional to me. I am still wondering why SPM doesn't automatically pick the correct product like it seems to suggest in its docs.

To better understand the specific issue you’re encountering, could you provide a minimal reproducible example or a more detailed explanation of the problem? That way we can rule out that this is a bug in SPM or potentially a misconfiguration on your end and make sure that this is in fact something we need to change on the library end. 🙏

dnl-moh:
Closed for inactivity.

PR #268, opened 2022-01-28
Add ECDH support for JWE Key Management

anthony-dedieu-ariadnext:
Add EC support for JWE Key Management

ECDH-ES - Elliptic Curve Diffie-Hellman Ephemeral Static key agreement using Concat KDF
ECDH_ES_A128KW - ECDH-ES using Concat KDF and CEK wrapped with "A128KW"
ECDH_ES_A192KW - ECDH-ES using Concat KDF and CEK wrapped with "A192KW"
ECDH_ES_A256KW - ECDH-ES using Concat KDF and CEK wrapped with "A256KW"

This pull request is a continuation of #245 which has been closed a while ago due to inactivity.

Tests have been fixed, TODOs have been resolved.

PR #304, opened 2023-12-29
Add support for PBES2

tobihagemann:
This PR builds upon PR #268.

You can find the actual added support for PBES2 in de60619e9a8182861da84d5602a69e6e142e8544.

### DIFF
--- a/JOSESwift/Sources/AlgorithmExtensions.swift
+++ b/JOSESwift/Sources/AlgorithmExtensions.swift
@@ -22,7 +22,6 @@
 //
 
 import Foundation
-import CommonCrypto
 
 extension ContentEncryptionAlgorithm {
 
@@ -84,11 +83,11 @@ extension KeyManagementAlgorithm {
     var keyLength: Int {
         switch self {
         case .PBES2_HS256_A128KW:
-            return kCCKeySizeAES128
+            return 16
         case .PBES2_HS384_A192KW:
-            return kCCKeySizeAES192
+            return 24
         case .PBES2_HS512_A256KW:
-            return kCCKeySizeAES256
+            return 32
         default:
             return 0
         }
--- a/JOSESwift/Sources/AlgorithmExtensions.swift
+++ b/JOSESwift/Sources/AlgorithmExtensions.swift
@@ -80,7 +80,7 @@ extension KeyManagementAlgorithm {
         }
     }
 
-    var keyLength: Int {
+    var derivedKeyLength: Int? {
         switch self {
         case .PBES2_HS256_A128KW:
             return 16
@@ -89,7 +89,7 @@ extension KeyManagementAlgorithm {
         case .PBES2_HS512_A256KW:
             return 32
         default:
-            return 0
+            return nil
         }
     }
 }
--- a/JOSESwift/Sources/AlgorithmExtensions.swift
+++ b/JOSESwift/Sources/AlgorithmExtensions.swift
@@ -22,6 +22,7 @@
 //
 
 import Foundation
+import CommonCrypto
 
 extension ContentEncryptionAlgorithm {
 
@@ -62,6 +63,34 @@ extension SignatureAlgorithm {
             return .SHA512
         default:
             return nil
+        }
+    }
+}
+
+extension KeyManagementAlgorithm {
+    var hmacAlgorithm: HMACAlgorithm? {
+        switch self {
+        case .PBES2_HS256_A128KW:
+            return .SHA256
+        case .PBES2_HS384_A192KW:
+            return .SHA384
+        case .PBES2_HS512_A256KW:
+            return .SHA512
+        default:
+            return nil
+        }
+    }
+
+    var keyLength: Int {
+        switch self {
+        case .PBES2_HS256_A128KW:
+            return kCCKeySizeAES128
+        case .PBES2_HS384_A192KW:
+            return kCCKeySizeAES192
+        case .PBES2_HS512_A256KW:
+            return kCCKeySizeAES256
+        default:
+            return 0
         }
     }
 }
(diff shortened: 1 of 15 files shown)


JOSESwift/Sources/PBES2KeyEncryptionMode.swift
```diff
@@ -0,0 +1,77 @@
...
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//  ---------------------------------------------------------------------------
+//
+
+import Foundation
+
+struct PBES2KeyEncryptionMode {
+    typealias KeyType = String
```
daniel-moh, line 27: From [RFC-7518](https://datatracker.ietf.org/doc/html/rfc7518#section-4.8):

> The PBES2 password input is an
   octet sequence; if the password to be used is represented as a text
   string rather than an octet sequence, the UTF-8 encoding of the text
   string MUST be used as the octet sequence.

I guess we had to choose to either have `String` or `Data` as a data type here. I'd love to hear the reasoning behind choosing `String` in this case. Not saying one is better than the other, just genuinely interested in the reasoning.

daniel-moh:
@tobihagemann thanks for opening this PR! One step closer to having all the algorithm checkboxes ticked. This looks very good already! 👏  There's a few minor changes that could improve readability and flow in some places imo. But since it took us so long to get back with a review, I don't want to bother you with these nits and will just push some changes to your branch to fix them. Hope this is ok with you.

tobihagemann:
@daniel-moh That's totally fine, thank you very much! :)

JOSESwift/Sources/AlgorithmExtensions.swift
```diff
@@ -65,3 +65,31 @@ extension SignatureAlgorithm {
...
+        case .PBES2_HS384_A192KW:
+            return .SHA384
+        case .PBES2_HS512_A256KW:
+            return .SHA512
+        default:
+            return nil
+        }
+    }
+
+    var keyLength: Int {
```
daniel-moh, line 83:
```suggestion
    var derivedKeyLength: Int {
```

JOSESwift/Sources/CryptoImplementation/PBES2.swift
```diff
@@ -0,0 +1,91 @@
...
+
+import Foundation
+import CommonCrypto
+
+internal enum PBES2Error: Error {
+    case unknownOrUnsupportedAlgorithm
+}
+
+internal struct PBES2 {
+    static let defaultIterationCount = 1_000_000
```
daniel-moh, line 32: A million iterations feels arbitrary. Is there any reason for picking exactly this number? If not we could go with the [recommended minimum](https://datatracker.ietf.org/doc/html/rfc7518#section-4.8.1.2) of `1_000`. thoughts?

JOSESwift/Sources/PBES2KeyEncryptionMode.swift
```diff
@@ -0,0 +1,77 @@
...
+        self.keyManagementAlgorithm = keyManagementAlgorithm
+        self.contentEncryptionAlgorithm = contentEncryptionAlgorithm
+        self.password = password
+    }
+}
+
+extension PBES2KeyEncryptionMode: EncryptionKeyManagementMode {
+    func determineContentEncryptionKey(for header: JWEHeader) throws -> Data {
+        var updatedHeader = header
+        let saltInput = try SecureRandom.generate(count: 16)
```
daniel-moh, line 47: The [recommended minimum is 8 octets](https://datatracker.ietf.org/doc/html/rfc7518#section-4.8.1.1). 16 feels a bit arbitrary, should we switch this to 8 (and ideally give the user the choice to specify this length themselves, see my other comment).

daniel-moh (dismissed):
Looks good! I pushed the following updates:

- Make the derived key length default value for unsupported algorithm combinations `nil`: https://github.com/airsidemobile/JOSESwift/pull/304/commits/edd4bbfab7a0b24a71c33c972cf12b74017d82d7
- Improve naming of PBES2 params: https://github.com/airsidemobile/JOSESwift/pull/304/commits/56fb966cd5338a1d8575857eae51120c015fa1fb
- Make p2s-length and p2c customizable (+tests): https://github.com/airsidemobile/JOSESwift/pull/304/commits/c3d2975ba1c822591e4ffd8afd1c51647945cca3

Let's merge this and include it in the next release 🚀  Thanks again @tobihagemann!

daniel-moh (approved):

JOSESwift/Sources/CryptoImplementation/PBES2.swift
tobihagemann: It has already been merged, but I wanted to mention that it's probably better to set a "secure" default than to set a minimum, defined in 2015.

1 million is probably "arbitrary", but OWASP recommends iteration counts between 210k and 1.3 million, see: https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html#pbkdf2

1000 is far too low for today's standards. I would argue that this wouldn't be a good "default" value.

JOSESwift/Sources/PBES2KeyEncryptionMode.swift
tobihagemann: I guess, there is no right or wrong here. Giving the user the choice is probably best. Not sure if 16 is still a better "default", see discussions like: https://security.stackexchange.com/questions/110084/parameters-for-pbkdf2-for-password-hashing

daniel-moh: You're right. I went with 8 as default (as it is the only reference to a number I found in the RFC) but opened it up for the user to specify whatever they want (as long as it's over 8 of course). Let me know what you think.

tobihagemann: Sounds great! 👍

PR #310, opened 2024-05-08
Update target platforms and fix tests

daniel-moh:
This is a maintenance PR to get the project one step closer to a state where we can start merging feature PRs again.

## Changes

- Adding watchOS, tvOS, and visionOS as officially supported platforms besides iOS and macOS
- Increasing the minimum deployment versions so that we can start using [Apple's CryptoKit](https://developer.apple.com/documentation/cryptokit/) framework
  - ⚠️ This will be a breaking change so our next release will need to be a major version bump. The deployment targets are still relatively low (e.g., iOS 13 was introduced in 2019) so this should be fine™️ 
- Fixes the tests that were relying on a specific order in the JSON serialization of JWKs by making sure JWKs get encoded in a deterministic way


haeser (approved):
Nice. Looking good. 
Agree with the major version bump.

PR #312, opened 2024-05-10
Add support for public and private header parameters

daniel-moh:
Adds support for working with arbitrary header parameters as per [RFC-7515, 4.2/3](https://www.rfc-editor.org/rfc/rfc7515#section-4.2) (JWS) and [RFC-1716-4.2/3](https://www.rfc-editor.org/rfc/rfc7516#section-4.2) (JWE).

## Special Considerations

- This API ensures that the dictionary used to access parameters and the underlying JSON data representation stay in sync. This avoids any situations where the parameters dictionary contains a value that cannot be represented in JSON and therefore the resulting JWS/E header does not.
- It is also guaranteed that required header parameters are not removed or altered via the new API. Required header parameters must be set when initializing the JWS/E header.

## Misc

Continuing what we started in #310, this adds deterministic JSON encoding (`.sortedKeys`) to the header JSON representation.

This outdates and closes #282.

### DIFF
--- a/JOSESwift/Sources/JOSEHeader.swift
+++ b/JOSESwift/Sources/JOSEHeader.swift
@@ -48,7 +48,7 @@ extension JOSEHeader {
             throw JOSESwiftError.invalidHeaderParameterValue
         }
 
-        guard JSONSerialization.isValidJSONObject([parameterName: value]) else{
+        guard JSONSerialization.isValidJSONObject([parameterName: value]) else {
             throw JOSESwiftError.invalidHeaderParameterValue
         }
 
--- a/JOSESwift/Sources/JOSEHeader.swift
+++ b/JOSESwift/Sources/JOSEHeader.swift
@@ -43,7 +43,7 @@ protocol JOSEHeader: DataConvertible, CommonHeaderParameterSpace {
 }
 
 extension JOSEHeader {
-    mutating func set(_ value: Any, forParameter parameterName: String) throws {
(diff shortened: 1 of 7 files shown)


haeser (approved):
👏🏼 very nice and clean approach that fits neatly.

tobihagemann:
Hey @daniel-moh, I finally found some time and wanted to migrate from our fork to the latest and greatest official version 3.0.0. 😉 However, I noticed something: When trying to access a header parameter via `header.get(parameter:)`, I'm getting the error: `'get' is inaccessible due to 'internal' protection level`.

Maybe the `JOSEHeader` protocol and/or the extensions have to be `public`? 🤔

Sorry for being lazy and not opening an issue. 🙈

jking-bit:
> Hey @daniel-moh, I finally found some time and wanted to migrate from our fork to the latest and greatest official version 3.0.0. 😉 However, I noticed something: When trying to access a header parameter via `header.get(parameter:)`, I'm getting the error: `'get' is inaccessible due to 'internal' protection level`.
> 
> Maybe the `JOSEHeader` protocol and/or the extensions have to be `public`? 🤔
> 
> Sorry for being lazy and not opening an issue. 🙈

Similar issue here. Have a `JWSHeader` but both `get(parameter parameterName: String)` and `set(_ value: Any, forParameter parameterName: String)` give a `is inaccessible due to 'internal' protection level`. 
@tobihagemann suggestion to pass `JOSEHeader` might be the solution

PR #341, opened 2024-08-23
Add support for A192CBC-HS384 and A192GCM

daniel-moh:
_🍇 Some low hanging fruits:_ Adds support for [A192CBC-HS384](https://datatracker.ietf.org/doc/html/rfc7518#section-5.2.4) and [A192GCM](https://datatracker.ietf.org/doc/html/rfc7518#section-5.3).

After this, the only algorithms still missing are the A*GCMKW variants. 🎉 

Screenshot 2024-08-23 at 20 29 37


haeser (approved):
nice harvest 😁

PR #353, opened 2024-09-04
Setup linters and reporting to sonarcloud

daniel-moh:
Sets up SwiftFormat and SwiftLint as version controlled build tools to report:

- inline in Xcode
- in SonarCloud

To keep this PR readable, both linters are currently configured to not report anything. We can configure rules and do a bulk format of the codebase in a next step.


daniel-moh:
Screenshot 2024-09-05 at 22 52 00

Reporting works for both SwiftFormat and SwiftLint. 🎉

haeser (approved):
🧾

